PR #19, opened 2025-04-30
Feature/organizations

K-Krut:


backend/dataforce_studio/models/organization.py
```diff
@@ -0,0 +1,58 @@
+import datetime
+import uuid
+from enum import Enum as Enumeration
```
Niek23:
```suggestion
from enum import StrEnum
```

backend/dataforce_studio/models/organization.py
```diff
@@ -0,0 +1,58 @@
...
+from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
+
+
+class OrgRole(str, Enumeration):
+    OWNER = "owner"
+    ADMIN = "admin"
+    MEMBER = "member"
+
+
+class DBOrganization(DeclarativeBase):
```
Niek23: Should inherit from the same Base as User model

PR #22, opened 2025-05-08
Feature/organizations adding user

K-Krut:


backend/dataforce_studio/api/organization.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+    try:
+        return await organization_handler.accept_invite(invite_id, request.user.id)
+    except AuthError as e:
+        raise handle_auth_error(e) from e
+
+
+@organization_router.post("/invite/reject", status_code=status.HTTP_204_NO_CONTENT)
+async def reject_invite_to_organization(invite_id: uuid.UUID) -> None:
+    return await organization_handler.reject_invite(invite_id)
```
Niek23: Would you consider splitting org management and user invitations? 
It could be something like 
`/me/invitations`
`/me/invitations/{id}/accept`
`/me/invitations/{id}/reject` 

wdyt

backend/dataforce_studio/handlers/organizations.py
```diff
@@ -0,0 +1,81 @@
...
+
+    __members_limit = 10
+
+    async def check_org_members_limit(
+        self, organization_id: uuid.UUID, num: int = 0
+    ) -> None:
+        members_count = await self.__user_repository.get_organization_members_count(
+            organization_id
+        )
+        members_count = members_count or 0
```
Niek23: Why is this needed?

backend/dataforce_studio/handlers/organizations.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if (members_count + num) >= self.__members_limit:
+            raise ValueError("Organization reached maximum number of users")
```
Niek23: Please add a custom exception for this

backend/dataforce_studio/models/orm/organization.py
```diff
@@ -61,3 +65,27 @@ def __repr__(self) -> str:
 
     def to_organization_member(self) -> OrganizationMember:
         return OrganizationMember.model_validate(self)
+
+
+class DBOrganizationInvite(TimestampMixin, Base):
```
Niek23:
```suggestion
class OrganizationInviteOrm(TimestampMixin, Base):
```

backend/dataforce_studio/models/orm/user.py
```diff
@@ -27,10 +27,8 @@ class UserOrm(TimestampMixin, Base):
     def __repr__(self) -> str:
         return f"User(email={self.email!r}, full_name={self.full_name!r}"
 
-    def to_user(
-        self,
-    ) -> User:
-        return User.model_validate(self)
+    def to_user(self) -> User:
+        return User.model_validate(self.__dict__)
```
Niek23: was the problem with using the object itself?

backend/dataforce_studio/models/user.py
```diff
@@ -28,6 +28,10 @@ class User(_UserBase, BaseOrmConfig):
     id: uuid.UUID
 
 
+class UserResponse(_UserBase):
```
Niek23, line 31: We don't want to expose hashed_password

backend/dataforce_studio/repositories/invites.py
```diff
@@ -0,0 +1,62 @@
...
+            invite = result.scalar_one_or_none()
+
+            if invite:
+                await session.delete(invite)
+            else:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND, detail="Invite not found"
+                )
+
+    async def get_invites_where(self, *where_conditions) -> list[OrganizationInvite]:
```
Niek23, line 37: Why do we want to pass where condition as an argument? I would avoid having anything repo related in the handler

backend/dataforce_studio/repositories/invites.py
```diff
@@ -0,0 +1,62 @@
...
+        async with self._get_session() as session, session.begin():
+            result = await session.execute(
+                select(DBOrganizationInvite).where(*where_conditions)
+            )
+            invites = result.scalars().all()
+
+            return [OrganizationInvite.model_validate(invite) for invite in invites]
```
Niek23: Can we add a method on the model for it?

backend/migrations/versions/003_add_org_tables_and_update_users.py
```diff
@@ -32,25 +33,36 @@ def upgrade() -> None:
         ),
     )
     op.add_column(
-        "users", sa.Column("updated_at", sa.DateTime(timezone=True), nullable=True)
+        "users",
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
```
Niek23, line 39: why do we need server default for it?

backend/migrations/versions/003_add_org_tables_and_update_users.py
```diff
@@ -32,25 +33,36 @@ def upgrade() -> None:
...
+            server_onupdate=sa.text("now()"),
+            nullable=True
+        )
     )
 
     now = datetime.datetime.now(datetime.UTC)
     conn = op.get_bind()
 
+    conn.execute(sa.text("UPDATE users SET updated_at = now() WHERE updated_at IS NULL"))
```
Niek23: updated_at can be nullable, this is not really needed

backend/migrations/versions/003_add_org_tables_and_update_users.py
```diff
@@ -32,25 +33,36 @@ def upgrade() -> None:
...
-                "UPDATE users SET id = :id, created_at = :now WHERE email = :email"
+                "UPDATE users SET id = :id, created_at = :now, updated_at = :now WHERE email = :email"
             ),
             {"id": str(uuid.uuid4()), "now": now, "email": email},
         )
 
     op.drop_constraint("users_pkey", "users", type_="primary")
     op.create_primary_key("users_pkey", "users", ["id"])
     op.alter_column("users", "created_at", nullable=False)
+    op.alter_column("users", "updated_at", nullable=False)
```
Niek23, line 68: updated_at should be nullable

backend/migrations/versions/003_add_org_tables_and_update_users.py
```diff
@@ -106,34 +118,36 @@ def upgrade() -> None:
...
 def downgrade() -> None:
-    op.drop_table("organization_members")
-    op.drop_table("organizations")
+    op.drop_constraint(None, "users", type_="unique")
     op.drop_column("users", "updated_at")
     op.drop_column("users", "created_at")
     op.drop_column("users", "id")
-    op.create_primary_key("users_pkey", "users", ["email"])
+    op.drop_table("organization_members")
+    op.drop_table("organizations")
```
Niek23: Why is it moved to the end?

backend/migrations/versions/003_add_org_tables_and_update_users.py
```diff
@@ -106,34 +118,36 @@ def upgrade() -> None:
...
 
 
 def downgrade() -> None:
-    op.drop_table("organization_members")
-    op.drop_table("organizations")
+    op.drop_constraint(None, "users", type_="unique")
     op.drop_column("users", "updated_at")
     op.drop_column("users", "created_at")
     op.drop_column("users", "id")
-    op.create_primary_key("users_pkey", "users", ["email"])
```
Niek23, line 139: why deleting this one?

backend/migrations/versions/004_add_organization_invites_table.py
```diff
@@ -0,0 +1,58 @@
...
+# orgrole_enum = ENUM("OWNER", "ADMIN", "MEMBER", name="orgrole", create_type=False)
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "organization_invites",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("email", sa.String(), nullable=False),
+        # sa.Column("role", orgrole_enum, nullable=False),
+        sa.Column('role', sa.Enum('OWNER', 'ADMIN', 'MEMBER', name='orgrole'), nullable=False),
```
Niek23: As it may change I would better just change this column to string

backend/dataforce_studio/repositories/invites.py
K-Krut: we can use this method to avoid duplicating code and then add methods to the repository with conditions, so that we could remove where conditions from handlers

PR #38, opened 2025-05-23
Feat/organizatin invites responses update

K-Krut:


backend/dataforce_studio/api/orbits.py
```diff
@@ -1,42 +1,36 @@
...
+from fastapi import APIRouter, status
 
 from dataforce_studio.handlers.orbits import OrbitHandler
-from dataforce_studio.infra.dependencies import is_user_authenticated
 from dataforce_studio.schemas.orbit import Orbit, OrbitCreate, OrbitDetails, OrbitUpdate
 
-orbits_router = APIRouter(
-    prefix="/orbits",
-    tags=["orbits"],
-    dependencies=[Depends(is_user_authenticated)],
```
Niek23, line 10: Looks like it was removed by mistake

Niek23: I see that it was moved to another place. 
As the api directory keeps growing I think we will have to organize it somehow better

backend/dataforce_studio/api/organization.py
```diff
@@ -1,35 +1,17 @@
...
-organization_router.include_router(invites_router)
-organization_router.include_router(members_router)
-organization_router.include_router(organization_orbits_router)
-organization_router.include_router(orbits_router)
-organization_router.include_router(orbit_router)
-organization_router.include_router(orbit_members_router)
-
-
-@organization_router.get("", response_model=list[OrganizationSwitcher])
+@organization_router.get("/all", response_model=list[OrganizationSwitcher])
```
Niek23: Doesn't look like rest endpoint. I think it can be something like me/organizations or users/me/organizations. We can also think of some other options

backend/dataforce_studio/api/user_invites.py
```diff
@@ -2,29 +2,26 @@
...
 from dataforce_studio.infra.dependencies import is_user_authenticated
-from dataforce_studio.schemas.organization import OrganizationInvite
+from dataforce_studio.schemas.organization import UserInvite
 
 user_invites_router = APIRouter(
-    prefix="/organization",
-    tags=["user-invites"],
+    prefix="/me/invitations",
     dependencies=[Depends(is_user_authenticated)],
 )
```
Niek23: Let's better take out all users/me operations from auth and make a new router /me or users/me. There we can also place organization endpoint

backend/dataforce_studio/repositories/invites.py
```diff
@@ -31,26 +34,35 @@ async def delete_organization_invite(self, invite_id: int) -> None:
...
+    ) -> Sequence[OrganizationInviteOrm]:
+        async with self._get_session() as session:
             result = await session.execute(
-                select(OrganizationInviteOrm).where(*where_conditions)
+                select(OrganizationInviteOrm).where(*where_conditions).options(*options)
             )
-            invites = result.scalars().all()
 
-            return OrganizationInviteOrm.to_invites_list(invites)
+            return result.scalars().all()
```
Niek23, line 45: why we don't use CrudMixin here?

backend/dataforce_studio/schemas/organization.py
```diff
@@ -25,12 +35,15 @@ class OrganizationInvite(BaseModel, BaseOrmConfig):
...
+    invited_by_user: UserOut | None = None
     created_at: datetime
 
 
+class UserInvite(OrganizationInvite):
+    organization: Organization | None = None
+
+
 class UpdateOrganizationMember(BaseModel):
-    id: int
```
Niek23, line 33: Why id is removed from here?

backend/dataforce_studio/repositories/invites.py
K-Krut: it will be refactored in another branch

backend/dataforce_studio/schemas/organization.py
K-Krut: not sure id needed also in schema when we request it route
what do you think?

PR #44, opened 2025-05-25
Feature/permissions

K-Krut:


backend/dataforce_studio/api/orbits.py
```diff
@@ -1,42 +1,74 @@
...
+@orbits_router.get("")
+async def get_organization_orbits(
+    request: Request, organization_id: int
+) -> list[Orbit]:
+    await permissions_handler.check_organization_permission(
+        organization_id,
+        request.user.id,
+        Resources.ORBIT,
+        ResourceAction.LIST,
+    )
```
Niek23: These checks should be within other handlers

PR #47, opened 2025-05-30
feature/organization-invites-emails

K-Krut:
- implemented method for sending email after invite send


backend/.env.example
```diff
@@ -10,4 +10,6 @@ CONFIRM_EMAIL_URL=https://dev-api.dataforce.studio/auth/confirm-email?confirmati
 
 CHANGE_PASSWORD_URL=http://localhost:5173/change-password?token=
 
+APP_AUTH_EMAIL_URL=https://app.dataforce.studio/sign-in
```
Niek23: What would happen if user is already logged in?

PR #60, opened 2025-06-13
feat: added fields to Orbit schema | added default sorting for orbit …

K-Krut:
…members

### DIFF
--- a/backend/dataforce_studio/handlers/permissions.py
+++ b/backend/dataforce_studio/handlers/permissions.py
@@ -84,7 +84,7 @@ class PermissionsHandler:
             action,
         )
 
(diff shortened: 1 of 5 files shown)


backend/dataforce_studio/handlers/permissions.py
```diff
@@ -77,15 +77,17 @@ async def check_orbit_action_access(
...
             action,
         )
-        await self.check_orbit_permission(
-            orbit_id,
-            user_id,
-            resource,
-            action,
-        )
+
+        if org_role != OrgRole.OWNER and org_role != OrgRole.ADMIN:
```
Niek23: org_role not in (..., ...)

PR #64, opened 2025-06-20
feat: added sending email when creating orbit with members or adding …

K-Krut:
feat: added sending email when creating orbit with members or adding new orbit member | added collections count to orbit response | fixed tests and errors

### DIFF
--- a/backend/dataforce_studio/handlers/orbits.py
+++ b/backend/dataforce_studio/handlers/orbits.py
@@ -107,17 +107,6 @@ class OrbitHandler:
         if not created_orbit:
             raise ServiceError("Some errors when creating")
 
-        if orbit.notify_by_email and created_orbit.members:
-            for member in created_orbit.members:
-                self.__email_handler.send_added_to_orbit_email(
-                    member.user.full_name
-                    if member.user and member.user.full_name
-                    else "",
-                    member.user.email if member.user and member.user.email else "",
-                    orbit.name if orbit else "",
-                    config.APP_EMAIL_URL,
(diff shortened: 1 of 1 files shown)


backend/dataforce_studio/handlers/emails.py
```diff
@@ -58,3 +59,19 @@ def send_organization_invite_email(
...
+        self, name: str, email: EmailStr, orbit: str, link: str
+    ) -> None:
+        message = Mail(
+            from_email=self.sender_email,
+            to_emails=str(email),
+        )
+        message.template_id = "d-f9f0925c82784b2f91a20c679ac532b3"
+        message.dynamic_template_data = {
+            "name": name,
+            "orbit": orbit or "",
```
Niek23, line 73: are there any cases when orbit name can be none?

backend/dataforce_studio/handlers/orbits.py
```diff
@@ -83,12 +86,25 @@ async def create_organization_orbit(
...
+        if orbit.notify_by_email and created_orbit.members:
+            for member in created_orbit.members:
+                self.__email_handler.send_added_to_orbit_email(
+                    member.user.full_name
+                    if member.user and member.user.full_name
+                    else "",
+                    member.user.email if member.user and member.user.email else "",
+                    orbit.name if orbit else "",
+                    config.APP_EMAIL_URL,
+                )
```
Niek23: How is sending happening? 
Do you think we may need to do it in background?

PR #69, opened 2025-06-24
feat: added limit on the number of organizations a user can belong to

K-Krut:


backend/dataforce_studio/infra/endpoint_responses.py
```diff
@@ -11,4 +11,8 @@
         "description": "Not enough rights for performing action for the resource",
         "content": {"application/json": {"example": {"detail": "Not enough rights"}}},
     },
+    500: {
+        "description": "Internal Server Error",
+        "content": {"application/json": {"example": "Internal Server Error"}},
+    },
```
Niek23: This should never happen, all 500 are the cases that we have to fix

PR #75, opened 2025-07-05
feat: added ml model manifest validation | fixed tests for ml models

K-Krut:


backend/dataforce_studio/schemas/ml_models.py
```diff
@@ -114,7 +159,7 @@ class MLModel(BaseModel, BaseOrmConfig):
     file_name: str
     description: str | None = None
     metrics: dict
-    manifest: dict
+    manifest: Manifest
     file_hash: str
     file_index: dict
```
OKUA1, line 164: dict[str, tuple[int, int]]

PR #83, opened 2025-07-09
feat: added buckets edit and delete | updated leave from org button -…

K-Krut:
… showed only when user has appropriate permission

### DIFF
--- a/frontend/src/components/organizations/registry/BucketSettings.vue
+++ b/frontend/src/components/organizations/registry/BucketSettings.vue
@@ -25,7 +25,7 @@
         </div>
       </div>
       <div class="bucket-form-wrapper">
-        <BucketForm :initial-data="initialData" :loading="loading" @submit="onFormSubmit" />
+        <BucketForm :initial-data="initialData" :loading="loading" :show-submit-button="false" @submit="onFormSubmit" />
       </div>
     </div>
     <template #footer>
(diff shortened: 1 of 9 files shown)


frontend/src/components/organizations/registry/BucketSettings.vue
```diff
@@ -0,0 +1,153 @@
...
+
+    const updatedBucket: any = {
+      id: props.bucket.id,
+      bucket_name: values.bucket_name,
+      endpoint: values.endpoint,
+      region: values.region,
+      secure: values.secure,
+      access_key: values.access_key,
+      secret_key: values.secret_key,
+    }
```
limanski-valera: const updateBucket = {
    ...values,
    id: props.bucket.id
}

No need to use any!

frontend/src/components/organizations/registry/BucketSettings.vue
```diff
@@ -0,0 +1,153 @@
...
+}
+
+.bucket-endpoint {
+  font-size: 12px;
+  color: var(--p-text-muted-color);
+}
+
+.bucket-form-wrapper :deep(button[type="submit"]) {
+  display: none;
+}
```
limanski-valera: Don't use ::deep - use form props to remove the button

limanski-valera (approved):

PR #93, opened 2025-07-22
feat: added endpoint for listing bucket urls for testing connection |…

K-Krut:
feat: added endpoint for listing bucket urls for testing connection | moved s3 logic to separate service


backend/dataforce_studio/services/s3_service.py
```diff
@@ -0,0 +1,76 @@
...
+
+from dataforce_studio.infra.db import engine
+from dataforce_studio.infra.exceptions import BucketConnectionError, NotFoundError
+from dataforce_studio.repositories.bucket_secrets import BucketSecretRepository
+from dataforce_studio.schemas.bucket_secrets import BucketSecret
+
+
+class S3Service:
+    def __init__(self) -> None:
+        self.__secret_repository = BucketSecretRepository(engine)
```
Niek23: Good idea to encapsulate the bucket logic with a dedicated client. However I'd avoid adding any repo related logic here. This class should remain focused on bucket operations. The required credentials should be passed into its methods directly rather than fetched internally

backend/dataforce_studio/services/s3_service.py
```diff
@@ -0,0 +1,62 @@
...
+    def _create_minio_client(self, secret: BucketSecret) -> Minio:
+        return Minio(
+            secret.endpoint,
+            access_key=secret.access_key,
+            secret_key=secret.secret_key,
+            session_token=secret.session_token,
+            secure=secret.secure if secret.secure is not None else True,
+            region=secret.region,
+            cert_check=secret.cert_check if secret.cert_check is not None else True,
+        )
```
Niek23, line 23: I would better move client creation to __init__ and store it as an instance attribute

backend/dataforce_studio/handlers/bucket_secrets.py
```diff
@@ -5,17 +5,39 @@
...
+    BucketSecretUrls,
 )
 from dataforce_studio.schemas.permissions import Action, Resource
+from dataforce_studio.services.s3_service import S3Service
 
 
 class BucketSecretHandler:
     __secret_repository = BucketSecretRepository(engine)
     __permissions_handler = PermissionsHandler()
+    __s3_service = S3Service()
```
Niek23: Let's better initiate where it is needed

backend/dataforce_studio/handlers/bucket_secrets.py
```diff
@@ -77,3 +99,18 @@ async def delete_bucket_secret(
...
+            organization_id, user_id, Resource.BUCKET_SECRET, Action.READ
+        )
+
+        object_name = "test_file"
+
+        return BucketSecretUrls(
+            presigned_url=await self._get_presigned_url(secret_id, object_name),
+            download_url=await self._get_download_url(secret_id, object_name),
+            delete_url=await self._get_delete_url(secret_id, object_name),
+        )
```
Niek23: Here we create minio client 3 times, should be only 1

backend/dataforce_studio/services/s3_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+    async def get_presigned_url(self, object_name: str) -> str:
```
Niek23: all the urls that we get are presigned, I think this one should be called get_upload_url

backend/dataforce_studio/services/s3_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+from minio import Minio
+
+from dataforce_studio.infra.exceptions import BucketConnectionError
+from dataforce_studio.schemas.bucket_secrets import BucketSecret
+
+
+class S3Service:
+    def __init__(self, secret: BucketSecret) -> None:
+        self._secret = secret
```
Niek23: we only need bucket name, the rest we don't need to keep here

backend/dataforce_studio/services/s3_service.py
```diff
@@ -7,8 +7,8 @@
 
 
 class S3Service:
-    def __init__(self, secret: BucketSecret) -> None:
-        self._secret = secret
+    def __init__(self, secret: BucketSecret, bucket_name: str) -> None:
+        self._bucket_name = bucket_name
```
Niek23: вот тут просто `self._bucket_name = secret.bucket_name`

PR #107, opened 2025-08-11
fix/DFS-247 added error handle for email sending

K-Krut:


backend/dataforce_studio/handlers/auth.py
```diff
@@ -99,25 +99,33 @@ def _verify_token(self, token: str) -> EmailStr:
...
-        )
+        try:
+            self.__emails_handler.send_activation_email(
+                user.email, confirmation_link, user.full_name
+            )
+        except Exception as error:
+            raise EmailDeliveryError(
+                "Error sending confirmation email.", 201
+            ) from error
+
```
Niek23: what is going to happen if sending email fails but the user is already created?

PR #113, opened 2025-08-25
feature/DFS-241: added logic for generation urls for models multipart upload

K-Krut:


backend/dataforce_studio/services/s3_service.py
```diff
@@ -24,6 +27,26 @@ def _create_minio_client(
...
 
+    @staticmethod
+    def _calculate_optimal_chunk_size(file_size: int) -> int:
+        if file_size <= 1073741824:  # 1gb
+            return 33554432  # 32 mb
+        if file_size <= 5368709120:  # 5gb
+            return 67108864  # 64 mb
+        if file_size <= 10737418240:  # 10gb
+            return 134217728  # 128 mb
+        return 268435456  # 256mb
```
OKUA1: S3 has a limit on a max number of parts. Very big models should have 5Gb parts.

backend/dataforce_studio/schemas/model_artifacts.py
```diff
@@ -130,6 +131,13 @@ class ModelArtifactCreate(BaseModel):
     tags: list[str] | None = None
     status: ModelArtifactStatus = ModelArtifactStatus.PENDING_UPLOAD
 
+    @field_validator("size")
+    @classmethod
+    def validate_model_size(cls, value: int) -> int:
+        if value > 5497558138880:
```
Niek23: Should be const

Niek23 (approved):

PR #118, opened 2025-09-03
feat: added description and dynamic status for satellites | added end…

Niek23:
…points for update satellite and regenerate api_key


Niek23:
@codex

PR #124, opened 2025-09-07
feature/DFS-262: add satellite logic

Niek23:


satellite/model_server/handlers/model_handler.py
```diff
@@ -0,0 +1,269 @@
...
+from .file_handler import FileHandler
+
+logger = logging.getLogger(__name__)
+
+
+class ModelHandler:
+    def __init__(self, url: str | None = None) -> None:
+        self._model_url = os.getenv("MODEL_ARTIFACT_URL") if url is None else url
+        self._file_handler = FileHandler()
+        self._cached_models = {}  # url -> local_path mapping
```
OKUA1: We only have one model, so no need to track cache via dict.

satellite/model_server/handlers/model_handler.py
```diff
@@ -0,0 +1,269 @@
...
+        self._request_model_schema = None
+        self.model_envs = None
+        self.conda_worker = None
+
+        try:
+            self.model_path = self._get_model_path()
```
OKUA1: If the model was already downloaded in the previous container run, it should not be downloaded again.

satellite/agent/handlers/model_server_handler.py
```diff
@@ -0,0 +1,140 @@
...
+                dep for dep in deployments_db if dep.get("status", "") == "active"
+            ]
+
+            logger.info(
+                f"[active_deployments_db] {[d.get('id', '') for d in active_deployments_db]}"
+            )
+            for dep in active_deployments_db:
+                async with ModelServerClient() as client:
+                    health_ok = await client.is_healthy(dep["id"])
+                if health_ok:
```
OKUA1, line 76: else redeploy ?

PR #305, opened 2026-01-22
feat/LM-227 s3 compatible server

K-Krut:


sdk/python/luml/s3_proxy/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+        prog="s3-proxy",
+        description="S3-compatible proxy server for local file storage",
+    )
+    parser.add_argument("--port", type=int, default=9000, help="Port to bind to")
+    parser.add_argument(
+        "--storage", default="./s3_storage", help="Storage root directory"
+    )
+    parser.add_argument("--access-key", help="AWS access key for authentication")
+    parser.add_argument("--secret-key", help="AWS secret key for authentication")
+    parser.add_argument("--cors", action="store_true", help="Enable CORS support")
```
Niek23: Are there use cases when it should be modified by the user? Or we can simply hardcode some values?

sdk/python/luml/s3_proxy/s3proxy.py
```diff
@@ -0,0 +1,512 @@
...
+) -> None:
+    S3ProxyHandler.STORAGE_ROOT = storage_root
+    S3ProxyHandler.DEBUG = debug
+
+    if access_key and secret_key:
+        S3ProxyHandler.CREDENTIALS = {access_key: secret_key}
+    else:
+        S3ProxyHandler.CREDENTIALS = {}
+
+    S3ProxyHandler.CORS_ENABLED = cors_enabled
```
Niek23: Not sure why we need to keep it in class variable and assign this way, let's better just pass all these things in init method

sdk/python/pyproject.toml
```diff
@@ -17,6 +17,7 @@ api = [
 
 [project.scripts]
 generate-docs = "generate_docs:main"
+s3-proxy = "luml.s3_proxy:main"
```
Niek23: Please also create a ticket to update documentation, so that we don't forget about it

olha-poliuliakkh (approved):

PR #472, opened 2026-04-15
fix/LM-558: disable simultaneous opening of org/orbits

allmok:


limanski-valera (changes requested):
Clean the component LayoutHeader

limanski-valera (approved):

PR #488, opened 2026-04-22
refactor: reorganized SQLite backend implementation

K-Krut:


sdk/python/sdk/luml/experiments/backends/sqlite_backend/_sqlite_attachments.py
```diff
@@ -0,0 +1,205 @@
+# flake8: noqa: E501
+import uuid
+
+from sdk.luml.artifacts._base import _BaseFile
+from sdk.luml.experiments.backends.data_types import (
+    AttachmentRecord,
+    FileNode,
+)
+from sdk.luml.experiments.backends.sqlite_backend._sqlite_base import _SQLiteBase
+from sdk.luml.utils.tar import create_and_index_tar
```
Niek23: these imports are only working locally
should be without `sdk.`

sdk/python/sdk/luml/experiments/backends/sqlite_backend/_sqlite_experiments.py
```diff
@@ -0,0 +1,1260 @@
...
+        values.append(experiment_id)
+        cursor.execute(
+            f"UPDATE experiments SET {', '.join(fields)} WHERE id = ?",  # noqa: S608
+            values,
+        )
+        conn.commit()
+
+        cursor.execute(
+            "SELECT id, name, created_at, status, tags, duration, description, group_id"
+            "FROM experiments WHERE id = ?",
```
Niek23: space is missing

Niek23:
Please also add a new GitHub workflow for running tests/lints

codie-dfs:
codie started reviewing commit `54be3b7` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `54be3b7` (requested by @OKUA1)

I've dispatched a deep faithfulness comparison of the moved method bodies and am waiting for it to complete before finalizing. So far my structural review has found a critical, import-breaking bug and a minor duplication. I'll incorporate the body-comparison results when the subagent returns.
Waiting for the faithfulness comparison to complete.
All findings verified. Here is my review.

---

## Summary

This PR splits the 4182-line `sqlite.py` monolith into a `sqlite_backend/` package of mixins (`SQLiteExperimentMixin`, `SQLiteTraceMixin`, etc.) composed into `SQLiteBackend`, plus lint-only touch-ups to `s3proxy.py`, `pyproject.toml`, and two tests. The split is faithful for the large majority of methods, but it introduces one import error that makes the backend (and the whole test suite) unimportable, two hard `IndexError` crashes from dropped SQL columns, and silent `source` data-loss across every experiment read.

## Correctness bugs

### 1. Critical — backend is unimportable (wrong `_search_utils` module path)
`sdk/python/sdk/luml/experiments/backends/sqlite_backend/_sqlite_evals.py:18` and `sdk/python/sdk/luml/experiments/backends/sqlite_backend/_sqlite_traces.py:21` import:
```python
from luml.experiments.backends.sqlite_backend._search_utils import ...
```
but `_search_utils.py` lives at `luml/experiments/backends/_search_utils.py` — there is no `_search_utils.py` inside `sqlite_backend/` (and `sqlite_backend/__init__.py` is empty). Both mixins are bases of `SQLiteBackend`, and `sdk/python/sdk/luml/experiments/backends/__init__.py:2` and `sdk/python/sdk/tests/conftest.py:8` import `SQLiteBackend` at module load — so `import luml.experiments.backends` and the entire test suite raise `ModuleNotFoundError`. `_sqlite_experiments.py:14` already uses the correct path (`backends._search_utils`); fix the other two to match.

### 2. Crash — model listing drops SQL columns → `IndexError` (also breaks experiment pagination)
`_sqlite_models.py:269` (`list_batch_experiments_models`) and `_sqlite_models.py:323` (`list_experiment_models`) select only 7 columns:
```sql
SELECT id, name, created_at, tags, path, size, experiment_id FROM models ...
```
but `_row_to_model` (`_sqlite_models.py:27-28`) reads `row[7]` (source) and `row[8]` (description). Any returned row raises `IndexError`. The old queries selected all 9 columns, and the sibling `get_models`/`_fetch_model` still do (`_sqlite_models.py:33`), so this is an inconsistent regression, not a schema change. Blast radius extends beyond these two methods: `_build_experiments_page` calls `list_batch_experiments_models` (`_sqlite_experiments.py:78`), so `list_group_experiments_pagination` / `list_groups_experiments_pagination` also crash whenever any listed experiment has a model. Re-add `, source, description` to both SELECTs.

### 3. Silent data loss — `source` dropped from every experiment read
The `source` column was stripped from the SELECT and from the `Experiment(...)` construction in four places, so `Experiment.source` (`data_types.py:70`) now always returns `None` even though the column is still written:
- `list_experiments` — `_sqlite_experiments.py:865` / constructor `872-883` (old read `source=row[10]`)
- `get_experiment` — `_sqlite_experiments.py:966-967` / constructor `976-988` (old read `source=row[11]`)
- `update_experiment` — SELECTs at `_sqlite_experiments.py:1058` and `1084`, constructors `1065-1074` / `1090-1099` (old read `source=row[8]`)
- `_build_experiments_page` via `_EXPERIMENT_COLUMNS` — `_sqlite_base.py:146-157` lost its final `"source"` entry, and `_sqlite_experiments.py:93-107` dropped `source=e.get("source")`

This is the same accidental column-stripping as finding #2 (here it fails silently because `Experiment.source` defaults to `None`). Restore the column and the assignment in each.

## Smaller problems

### 4. Behavior change bundled into a "pure move": attachment hardening
`log_attachment` (`_sqlite_attachments.py:47-55`) adds two new `ValueError`s (`bytes` + `binary=False`, `str` + `binary=True`), changes the open mode to key on `isinstance(data, bytes)` instead of the `binary` flag, and routes both `log_attachment:52` and `get_attachment:109` through a new `_safe_attachment_path` traversal guard that raises on names escaping the attachments dir. The traversal check is a genuine improvement, but these are new exceptions/behaviors that don't belong silently inside a refactor — call them out and cover them with tests (see below).

### 5. Duplicate `export_attachments`
Defined identically in both `_sqlite_experiments.py:1204` and `_sqlite_attachments.py:198`. MRO makes the `SQLiteExperimentMixin` copy win, so there's no behavioral impact, but the attachments-mixin copy is dead. Keep one.

## Missing test coverage

No tests were added, and the existing SQLite tests can't run at all (finding #1 breaks `conftest.py` import), so none of the above was caught. Before merge, add at minimum:
- an import smoke test for `SQLiteBackend` (guards #1),
- a regression test for `list_experiment_models` and a paginated experiment listing where a model exists (guards #2),
- assertions that `source` round-trips through `get_experiment` / `list_experiments` / `update_experiment` (guards #3),
- tests for the new attachment validation and path-traversal rejection (covers #4).

Note: the other mixins (base connection pool, traces, evals, groups, annotations, pagination) were spot-checked and appear to be faithful moves.

PR #531, opened 2026-05-15
feat: spec for dataset card

K-Krut:


codie-dfs:
codie started implementing this spec on branch `codie/feat/codie-impl-datasets-card`.

codie-dfs:
codie finished. PR opened from `codie/feat/codie-impl-datasets-card` into `feat/codie-impl-datasets-card`.

PR #585, opened 2026-06-11
feat/LM-630: add tracks frontend

limanski-valera:


K-Krut (commented):
Image

stages field does not match design

K-Krut (commented):
Image

by default, without filters applied, we show all artifacts from orbit

---------------------------------------------------

Image

Image

artifact cards incorrect 
---------------------------------------------------

Image

artifacts are not filtered by track type. artifacts must only be the same type as the defined track type

---------------------------------------------------

Image

artifact id isnt clickable

K-Krut (approved):

PR #589, opened 2026-06-16
fix/LM-318: show active deployments when deleting a model

limanski-valera:


K-Krut (changes requested):
Image
please fix ui for this warning

Also, when we try to delete a model with settings, we don't see this warning; we just see the error 'Failed to delete artifact "name".'

And please resolve conflicts

K-Krut (approved):

PR #596, opened 2026-06-26
fix: traces yaml text view

limanski-valera:


codie-dfs:
codie started reviewing commit `3978660` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `3978660` (requested by @OKUA1)

## Summary

This PR replaces a hand-rolled `jsonToYaml` with `js-yaml`'s `dump()` in the two parallel `UiMultiTypeText` components (`frontend` and `extras/js/packages/experiments`), and makes the YAML view first attempt to parse a string prop as JSON before dumping it. It also exports the previously-private `tryParseJson` and tweaks `.yaml-body` CSS. Switching object→YAML rendering to a real library is a good correctness improvement over the old ad-hoc serializer.

## Dependency concern (please verify before merging)

**`js-yaml` is pinned to `^5.1.0`, which looks wrong.** In `frontend/package.json` and `extras/js/packages/experiments/package.json`:

- The added types are `@types/js-yaml@4.0.9`, i.e. the **4.x** API — the same API the code uses (`dump(obj, { indent, lineWidth })`). Because the types are 4.x, a genuine 5.x runtime API change would **not** be caught by typecheck.
- The lockfile already resolves `js-yaml@4.1.1` at the root (`package-lock.json:11597`, a normal 4.x with `bin/js-yaml.js`). The new `^5.1.0` instead pulls a **separate, duplicated** nested copy (`package-lock.json:685` / `:494`) resolved to `5.1.0` with a different bin (`bin/js-yaml.mjs`).

The mainline `js-yaml` package's well-known latest is 4.x, so `^5.1.0` is a surprising jump. I couldn't confirm `5.1.0` against the npm registry in this environment, so please double-check it's an intentional, legitimate version — a wrong version pin is both a supply-chain risk and can break `npm ci`. Recommended fix: pin `^4.1.1` to match `@types/js-yaml@4.0.9` and dedupe with the copy already in the tree. (Minor: in both files `js-yaml` is appended out of alphabetical order in the `dependencies` block — e.g. `extras/js/packages/experiments/package.json:67-68`.)

## Minor

- **`lineWidth: 160` will fold long values** (`frontend/src/helpers/helpers.ts:194`, `extras/js/packages/experiments/src/helpers/texts.ts:13`). `js-yaml` inserts real line breaks into scalars longer than the limit. That works against the new `.yaml-body` styling (`white-space: pre` + the parent's `overflow-x: auto`), whose intent appears to be "don't wrap, scroll horizontally" — long string values now get hard-wrapped in the data instead. If verbatim display + horizontal scroll is the goal, `lineWidth: -1` disables folding. The old serializer never folded, so this is a behavior change for long values.

## Missing test coverage

There are no tests for `jsonToYaml`, `tryParseJson`, or the new `yamlText` path, and the helper's output format changed materially (quoting, trailing newline, folding, `null`/boolean rendering). Worth adding unit tests for the new branch in `yamlText` (`frontend/src/components/ui/multi-type-text/UiMultiTypeText.vue:82-85`):

- JSON string (`'{"a":1}'`) → parsed and dumped as YAML.
- Non-JSON string (`'hello: world'`) → passed through unchanged (verifying `?? props.text` fallback).
- Falsy-but-valid JSON values (`'false'`, `'0'`) → correctly kept, confirming `??` was chosen over `||`.
- Object prop → dumped directly.

olha-poliuliakkh (approved):

PR #597, opened 2026-06-28
feat: add lineage UI

limanski-valera:


codie-dfs:
codie started reviewing commit `51d62f8` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `51d62f8` (requested by @OKUA1)

I have enough to write the review. Here is my assessment.

---

## Summary

This PR adds a lineage UI for artifacts, built on `@vue-flow/core`: a new route/page (`LineageView.vue`), a set of `components/lineage/*` components, a `useLineageStore` Pinia store managing nodes/edges/undo-history, and supporting changes to the shared `useArtifactsList` hook, `confirm.ts`, and vue-flow styling. The feature is largely UI scaffolding — persistence (`save()`) is stubbed, initial data is never loaded, and a couple of components are placeholders — so much of it reads as WIP. The findings below are logic issues that are real in the current code, plus a regression in a shared hook that affects an existing component.

## Correctness

### 1. Undo / Reset-positions history tracking is broken by a debounce/`nextTick` race — `frontend/src/stores/lineage/index.ts`

The `isRestoring` guard (`:33`) is meant to stop restore operations from being recorded as new history, but it can never do so. `goBack()` (`:46-52`) and `resetPositions()` (`:119-125`) set `isRestoring = true`, call `setNodes/setEdges`, then reset `isRestoring = false` inside `nextTick(...)`. But the change handler is debounced by 200ms (`:147`) and registered on `onNodesChange`/`onEdgesChange` (`:149`). `nextTick` is a microtask that resolves long before the 200ms `setTimeout`, so by the time `onStateChanged` (`:111`) actually runs, `isRestoring` is already `false` and `if (isRestoring) return` never fires.

Consequences:
- After `goBack()`, the restored state is re-appended to `history` (`:113`), so `history.length` doesn't shrink — the Undo/Save/Reset buttons (all gated on `history.length === 0`) stay enabled and undo doesn't progress cleanly.
- `resetPositions()` clears `history` (`:122`) but the deferred `onStateChanged` then pushes `prev` back. Since `resetPositions` never updates `prev`, `history` ends up holding the *pre-reset* (modified) state, so a subsequent Undo restores the wrong state.

The flag needs to stay set until after the debounced handler has run (e.g. reset it inside/after `onStateChanged`, or cancel the pending debounced call), not on `nextTick`.

### 2. Search now triggers a double `getInitialPage()` in `LinkArtifactCreator` — `frontend/src/hooks/useArtifactsList.ts:170`

Adding `searchQuery` to `watch([sortData, typesQuery, searchQuery], debouncedOnSortDataChange)` changes behavior for existing consumers of this shared hook. `LinkArtifactCreator.vue:151-154` already handles search manually: `setSearchQuery(query)` followed by `resetList()` (→ `getInitialPage()`). With the new watch, `setSearchQuery` *also* schedules `debouncedOnSortDataChange` → `onSortDataChange` → `getInitialPage()`. So each search fires two initial-page requests (~500ms apart) plus an extra `setList([])`, causing a redundant fetch and a list-clear flicker. Either remove the manual `resetList()` in `LinkArtifactCreator` or don't watch `searchQuery` in the hook.

The other consumers (`ArtifactsTable`, `ModelSelect`) don't call `setSearchQuery`, so they're unaffected — but the `getInitialPage` cursor change (`:66`, `push` → `= [response.cursor]`) and the `onLazyLoad` condition change (`:129`, `pageIndex * limit` → `filteredList.length`) also touch shared behavior and should be smoke-tested against those two tables.

## Smaller problems

- **Dead guard in `ArtifactSelector.vue`.** `isResetting` (`:95`) is read in three watcher guards (`:135`, `:143`, `:151`) but never assigned `true` anywhere, so the guards are no-ops. If the intent was to suppress the collection/type/search watchers during a programmatic reset, that logic is missing; otherwise remove the ref.

- **Placeholder copy shipped in `confirm.ts:250-256`.** `saveLineageChangesConfirmOptions` has `message: 'NEW TEXT NEDEED'` (also a typo for "NEEDED") and a `// TODO: Change message text`. It isn't imported anywhere, so it's currently dead placeholder code — wire it up with real text or drop it before merge.

- **`save()` is a stub — `stores/lineage/index.ts:128-138`.** It `setTimeout`s 1000ms, has the real `api.lineage.save(...)` commented out, and only mutates local refs. Nothing is persisted. Fine if this is intentionally WIP, but worth flagging since the "Save changes" button implies persistence.

- **`useVueFlow()` inside a Pinia store — `stores/lineage/index.ts:21`.** Every other `useVueFlow()` call in the repo is in component `<script setup>`; here it runs at store-setup. A Pinia store is a persistent singleton, so the destructured handles and the `onConnect`/`onNodesChange`/`onEdgesChange` registrations are bound once and never cleaned up, and `history`/`prev`/`initialNodes` leak across visits. Neither the store nor `<VueFlow>` in `LineageArea.vue` sets an explicit `id`, so please verify they resolve to the same flow instance, and add a reset on unmount so re-entering the page starts clean.

- **Placeholder components.** `LineageArtifactDetails.vue` is an empty `<div>`, and `ArtifactDetailsModal.vue` renders literal "Details" with a non-wired "View artifact" button. Presumably WIP.

- **Unrelated change:** `forceStageReassignConfirmOptions` was removed from `confirm.ts`. It appears unused elsewhere so it's safe, but it's unrelated to lineage and adds noise to the diff.

- **Trivial:** `LineageWrapper.vue:19` `const {} = useVueFlow()` is a no-op destructure and can be removed.

## Missing tests

No tests accompany ~1500 lines of new code. The most valuable to add, and the ones that would have caught the bugs above:
- `useLineageStore` history logic — record on change, `goBack`, `resetPositions`, and specifically that a restore does **not** grow `history` (finding #1 is directly testable with fake timers over the 200ms debounce).
- `useArtifactsList` — `filteredList` excluding `excludedArtifactIds`, the `onLazyLoad` trigger boundary, and that a `searchQuery` change results in a single reload (finding #2).

K-Krut (approved):

PR #602, opened 2026-07-02
feat: initial 2nd part

K-Krut:


codie-dfs:
codie started reviewing commit `899a344` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `899a344` (requested by @OKUA1)

I have enough to write the review. My verification established the key facts:

- **Part 1 is absent from the repo**: whole-repo search for `inference_events`, `monitoring_enabled`, `runtime_metrics`, `otel_traces`, `greptime`, `reference_profile`, `PSI`, `reconstruction_error`, `feature_drift`, `output_drift` returns nothing but `SPEC.md`. `satellite/docker-compose.yml` has only `model-image`/`agent` (no GreptimeDB/otel-collector). Neither `DeploymentOrm` nor the Satellite `Deployment` schema has a `monitoring_enabled` or profile field.
- The one grounded claim is the deploy flow reading manifest/schema (`satellite/agent/tasks/deploy.py` does load schemas).
- Algorithmic issues in the spec text itself (PSI zero-handling, multivariate 3σ calibration).

Here is my review:

---

## Summary

This PR adds one new file, `SPEC.md` — a design spec for the second slice of the Live Monitoring system: an SDK-generated **reference profile** (per-feature summaries, output summary, PCA profile) embedded in the model artifact, Satellite-side loading of that profile at deploy, and a per-Satellite **Monitoring Worker** that computes runtime health, data quality, univariate/output drift (PSI), and multivariate drift (PCA reconstruction error) into `monitoring_results` / `monitoring_alerts`. The document is clear, well-structured, and the task breakdown is coherent. The notes below are about the spec content itself; the largest concern is that the "part 1" foundation it builds on is not present in this repository.

## Findings

### 1. The "part 1" foundation the spec builds on isn't in this repo

`SPEC.md:82-96` states part 1 "is implemented and this slice builds directly on it," naming concrete infrastructure it will "reuse … and not re-add": GreptimeDB with an `inference_events` trace table, OpenTelemetry metric tables, `otel_traces`, a per-deployment `monitoring_enabled` flag, an advertised monitoring capability, and Agent/model-server telemetry collection. I could not find any of these in the checkout:

- `inference_events`, `monitoring_enabled`, `runtime_metrics`, `otel_traces`, `greptime` — no matches anywhere except `SPEC.md`.
- `satellite/docker-compose.yml` defines only `model-image` and `agent`; there is no GreptimeDB or otel-collector service.
- `backend/luml/models/deployment.py` (`DeploymentOrm`) and `satellite/agent/schemas/deployments.py` (`Deployment`) have no `monitoring_enabled` field and no reference-profile field.
- No inference-data collection (inputs/outputs/status/latency) exists to read from.

If part 1 lives in another repository or an unmerged branch, treat this as a note. But the spec asserts it as an implemented, in-repo dependency, and Tasks 2–6 are entirely predicated on it (attaching the profile to the deployment record, reading `inference_events`, reading the metric tables). Recommend restating these as prerequisites — and where they live — rather than "already exist … does not re-add them," so an implementer doesn't start Task 3 expecting a collection layer that isn't there. The one claim that does ground out today is the deploy flow reading the manifest/schema (`satellite/agent/tasks/deploy.py` loads schemas via `get_deployment_schemas`), so Task 2's "alongside the existing schema/manifest" is fine.

### 2. PSI definition only guards one of the two zero cases — `SPEC.md:224-226`

The formula `PSI = Σ (live_prop − ref_prop) · ln(live_prop / ref_prop)` is correct, but "a small epsilon to avoid division by zero" only covers `ref_prop = 0`. The term also diverges when `live_prop = 0`: `ln(0 / ref)` → −∞, multiplied by `(0 − ref) < 0`, gives **+∞**. Empty live bins are common with small windows, and the spec explicitly wants reference categories/bins with no live samples, plus unseen categories (`SPEC.md:224`), to contribute to the score. As written, a literal implementation yields `inf`/`NaN` PSI. The spec should floor/smooth **both** live and reference proportions by epsilon (and renormalize), not just the denominator.

### 3. Multivariate 3σ threshold can be mis-calibrated by chunk size — `SPEC.md:165`, `246-248`

The reference reconstruction-error `mean`/`std` are computed "over chunks of the training data" at packaging time (SDK), while the alert compares the **live window mean** error to `mean + 3·std`. Because the variance of a mean scales ~1/n, the 3σ rule is only calibrated if the reference `std` is the std of a chunk **mean** at the **same sample size as the live window**. Window size is a worker-side runtime config (the spec notes it's configurable) whereas chunk size is fixed at packaging — set in different places/times, and the spec never ties them together. If training chunks are larger than the live window, `std` is too small → chronic false alerts; if smaller, real drift is missed. Recommend specifying that the reference statistic is computed over chunks sized to the monitoring window (or otherwise reconciled with it).

### Smaller

- **`SPEC.md:202-203`** — the runtime metrics list includes both `error count` and `failed-inference count` as distinct metrics but never defines how they differ (both derive from `inference_events` status). Clarify the distinction or drop the duplicate.
- **`SPEC.md:269`** — the `metric_group` enum is `runtime, data_quality, feature_drift, output_drift, multivariate`; the last breaks the `*_drift` naming of its siblings. Minor, but results/dashboards key off these strings — consider `multivariate_drift` for consistency.
- **Data-quality alerts (`SPEC.md:214-215` region)** — `type-mismatch > 0%` and `unseen-category > 0%` warn on a single occurrence in a window, which will be noisy for high-cardinality or novel string inputs. Consider a minimum-count floor, or state that single-occurrence warnings are intended.

## Test coverage (per-task test bullets)

The task test bullets cover happy paths well but miss the edge cases the findings above concern:

- **PSI numerical stability** (Task 5): add an empty-live-bin window and an all-live-in-one-bin window (finding #2) and assert PSI stays finite.
- **Multivariate calibration** (Task 6): add a window whose size differs from the reference chunk size (finding #3) and assert the threshold still behaves.
- **Output drift for classification without scores** (Task 5): the tests mention shifted inputs/outputs but not the "no probabilities/scores produced" branch that the output summary and output drift describe.
- **`profile_status` surfacing** (Tasks 3–4): the missing/placeholder-profile path is tested for *skip* behavior, but not that `profile_status` is correctly set to "placeholder" in `monitoring_results` when a placeholder profile is used.

codie-dfs:
codie started implementing this spec on branch `codie/feat/monitoring-metrics-processing`.

codie-dfs:
codie finished. PR opened from `codie/feat/monitoring-metrics-processing` into `feat/monitoring-metrics-processing`.

PR #604, opened 2026-07-03
fix/LM-636: tables bugs

limanski-valera:


codie-dfs (commented):
### codie review of `cc2b7d1` (requested by @OKUA1)

I've completed my investigation. Here is my review.

---

## Summary

This PR fixes eval-table bugs by introducing an `isEmpty()` helper and switching several ID-validation checks from truthy tests (`!value`) to `isEmpty(value)`, coercing IDs to strings with `String(...)` at call sites, and removing the `|| ''` fallbacks in `prepareEvalData` so `id`/`dataset_id`/`modelId` retain their real (possibly numeric or `null`) values. The rest is CSS tweaks. The data-flow change is coherent with the existing `hasRealId`/`__no_id_` sentinel handling, but the `isEmpty` migration is applied inconsistently, which reintroduces the exact class of bug it's meant to fix for one edge case.

## Correctness

**Partial `isEmpty` migration leaves falsy-but-valid IDs (e.g. `0`) broken in sibling code paths.** The whole point of `isEmpty` is that it differs from `!value` only for `0`/`false`/`NaN` — i.e. the codebase deliberately treats an ID of `0` as present (see `hooks/useEvalsTable.ts:36`, `hasRealId` = `item.id != null && item.id !== ''`). `selectedEval.evalId` is set from `data[0].id` (`store/evals/index.ts:42`), which after this PR's `safeParse` change can be the number `0`, and it flows into `AnnotationsView` as `props.evalId` (`TracesModel.vue:77`). But only some checks were migrated:

- `components/annotations/view/AnnotationsView.vue:98-116` — `openAddDialog` correctly uses `isEmpty`, so for `evalId === 0` it opens the **eval** add dialog. Yet in the same component:
  - `AnnotationsView.vue:131-137` — `addAnnotation` still does `if (props.evalId) … else if (props.spanId) … else throw`. For `evalId === 0`, submitting the dialog falls through to `throw new Error('Trace ID or span ID is required')` — the user opens an eval annotation dialog but cannot submit it.
  - `AnnotationsView.vue:66` (`currentAnnotations`) and `:8` (`:type="props.evalId ? 'eval' : 'span'"`) also use truthy checks, so an `evalId === 0` row misroutes to the span branch and won't display its eval annotations.
- `store/annotations.ts:63` (`addEvalAnnotation`) was migrated to `isEmpty`, but `store/annotations.ts:131` (`addSpanAnnotation`) still uses `!artifactId || !traceId || !spanId`.

Either `0` is a possible ID (then these sites are bugs) or it isn't (then the `isEmpty` refactor is unnecessary). Please apply the check consistently across all ID gates for the eval/span paths.

## Minor

- **Redundant/dead CSS declaration.** `components/evals/EvalsTable.vue:578` sets `color: inherit` and `:580` now adds `color: var(--p-primary-color)` in the same `.id-cell` block; the first is dead and overridden. Replace line 578 rather than duplicating.

- **`String()` coercion applied inconsistently in the same component.** `TracesModel.vue:187-188` wraps `selectedEval?.datasetId`/`evalId` in `String()` before calling `getEvalAnnotations`, but `showSpans` at `TracesModel.vue:160-161` passes the same `selectedEval` `datasetId`/`evalId` straight into `getEvalSpansTree(...: string)` unwrapped. If numeric IDs are the motivation for the coercion, this path has the same latent string/number mismatch.

- **`isEmpty(value: any)`** (`helpers/helpers.ts:146`) would be safer typed as `unknown`. Note it is not a type guard (`value is …`), so it does not narrow — which is why every call site still needs `String(...)`; that's fine, just be aware the casts are load-bearing.

## Missing tests

`isEmpty` is a pure function whose semantics are the crux of this fix, yet there's no test pinning them down — in particular that `0` and `false` are **not** empty while `''`/`null`/`undefined` are. The `@luml/experiments` package currently has no test runner configured (`package.json` has no `test` script), so this is a pre-existing gap, but given the subtle `isEmpty` vs `!value` distinction driving this PR, a unit test locking in `isEmpty(0) === false` / `isEmpty('') === true` would guard against future regressions and the inconsistencies noted above.

codie-dfs:
codie started reviewing commit `614b748` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `614b748` (requested by @OKUA1)

I have a complete understanding of the change. Let me write up the review.

---

## Summary

This PR fixes eval/traces table bugs (LM-636). The core fix is in `ExperimentSnapshotDatabaseProvider.prepareEvalData`, where `id`/`dataset_id` now use `?.toString() ?? ''` and `modelId` is passed through verbatim, instead of `safeParse(...) || ''` — which could turn numeric-looking IDs into numbers and corrupt JSON-parseable model IDs (e.g. `"0"`, `"null"`). A new `isEmpty` helper plus `String(...)` coercion is applied in the annotation flows to defend against numeric/zero IDs, alongside several CSS tweaks.

The central change is sound: because `selectedEval` is populated from `EvalsInfo.id`/`dataset_id` (`store/evals/index.ts:42`), the old `safeParse` path let numbers leak into fields typed as `string`, which is exactly what the `String(...)`/`isEmpty` guards in the components were added to absorb. The whole set of edits is internally coherent, and I found no correctness or security issues.

## Smaller problems

- **Duplicate/dead CSS declaration** — `extras/js/packages/experiments/src/components/evals/EvalsTable.vue:578` still sets `color: inherit;`, which is immediately overridden by the new `color: var(--p-primary-color);` at line 580. The `inherit` line is now dead; collapse to a single declaration.

- **Inconsistent validation migration** — `extras/js/packages/experiments/src/store/annotations.ts:131` (`addSpanAnnotation`) still uses `if (!artifactId || !traceId || !spanId)`, while the parallel `addEvalAnnotation` at line 63 was migrated to `isEmpty(...)`. If the numeric/zero-ID problem can affect span IDs too, the fix is incomplete on the span side; if not, the asymmetry is at least worth aligning. The same asymmetry exists in `AnnotationsView.vue`: `openAddDialog` uses `isEmpty`/`String`, but the routing checks it relies on (`props.evalId ? ... ` at lines 8, 66, 131) remain truthy checks — so a hypothetical `evalId` of `0` would route inconsistently between them.

- **Magic number in layout** — `extras/js/packages/experiments/src/components/evals/traces/inputs/TracesInputs.vue:74`, `height: calc(100% - 63px)`, hard-codes the sibling header height and will silently break if that header's height changes. A CSS variable or flex-based layout would be more robust, though this is cosmetic.

## Missing test coverage

No tests are included. The two pieces of pure logic here are the ideal regression-test targets: `isEmpty` (`helpers.ts:146`) and, more importantly, the `prepareEvalData` string coercion — a test asserting that a numeric row `id`/`dataset_id` (e.g. SQLite `INTEGER` `123`, and the `0` edge case) comes back as the string `"123"`/`"0"` would lock in the actual root-cause fix and prevent regression to the `safeParse` behavior.

That said, the `experiments` package currently has no test harness at all (no `test` script, no spec files, no vitest config), so this would require standing up test infrastructure rather than just adding a file — flagging it as a gap rather than a blocker.

K-Krut (approved):

PR #617, opened 2026-07-08
Feat/lm 632 time series

olha-poliuliakkh:


codie-dfs:
codie started reviewing commit `d052d32` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `d052d32` (requested by @OKUA1)

Confirmed — `onFutureChange` only updates `futureState`; the stale `rawForecast`/`rawFuture` drive the chart overlay, "Provided" points, and download. All findings are verified. Here is my review.

---

## Summary

This PR adds an end-to-end time-series forecasting feature: a SARIMAX auto-configuration engine in the Python webworker (`dfs_webworker/forecasting.py`) that JSON-serializes fitted pipelines into standalone `.luml` bundles, plus a full Vue flow (upload → column-role setup → train → evaluate/re-forecast), a registry model card, and the worker/route/store wiring to connect them. It's a large, carefully engineered change (~7k lines of real source and tests) with strong test coverage and evident attention to numerical edge cases and ApexCharts/pandas quirks. No high-severity or normal-use crashers were found; the notes below are edge cases, a packaging-hygiene issue, and test gaps.

## Correctness & edge cases

- **Non-ISO date columns bucket into the wrong period in non-UTC browsers** — `frontend/src/lib/data-processing/forecasting-setup.ts:31-38`. `toDate` uses `Date.parse` (local-time for non-ISO strings like `"02/01/2020"`), but `forecastDate`/`periodKey`/`formatUtcDate` (`:98-163`) read via `getUTC*`/`toISOString`. East/west of UTC, a non-ISO date near a period boundary shifts a period; for known-future models the frontend future-date grid then misaligns with the engine's `_generate_future_dates`, so re-forecast fails validation. The docstring's "all arithmetic is UTC … regardless of timezone" only holds for ISO-parsed inputs. `periodsBetween` day/week (`:85-88`) and the DatePicker-driven `end` (a local-midnight `Date`) share this root cause and can be off-by-one for exact-day selections.

- **Re-forecast horizon depends on the aggregated last row's arbitrary day-of-month** — `forecasting-setup.ts:66-73` (`monthsBetween`, via `periodsBetween` for month/quarter/year). Training data is period-aggregated, so the last row's day is arbitrary, yet the horizon uses `to.getUTCDate() > from.getUTCDate()`. The same selected end date yields a different horizon depending on that day: last history `2020-01-01` + end `2020-02-28` → horizon 2 (through Mar); `2020-01-31` + same end → horizon 1 (Feb only). This changes both the training `previewHorizon` and how many known-future rows the user must fill.

- **Re-forecast chart and download go stale after editing future values** — `frontend/src/components/express-tasks/forecasting/evaluate/index.vue:312-314` and `:359`. Only a change to `endDate` clears `rawForecast`/`rawFuture`; editing a cell in the future-values grid updates `futureState` but leaves the previous forecast in place. After running a known-future forecast and then changing an input value, the forecast line, the dotted "Provided" points (`:296-310`), and "Download predictions" (`:335`) all keep showing the old inputs with no staleness indication — the grid now disagrees with the chart. Consider also clearing the computed forecast when `futureState` changes.

- **An auxiliary column named exactly `<target>_lower`/`_upper` corrupts the target's interval** — `wasm/packages/dfs_webworker/dfs_webworker/forecasting.py:769-773` and `frontend/src/lib/data-processing/forecasting.ts:30-47`. `record[f"predicted_{col}"]` for an auto-forecast aux named `<target>_lower` overwrites the target's `predicted_<target>_lower` key in the same record (Python side), and the normalizer swallows it as the target bound (TS side). Pathological naming only, but the "disambiguation" the docstring claims doesn't cover an exact collision — a name-clash guard in `_validate_roles` would close it.

## Packaging / repo hygiene

- **36 committed build artifacts, and the `dfs_webworker` copy is stale** — `wasm/packages/dfs_webworker/build/lib/**` and `wasm/packages/promptopt/build/lib/**`. `wasm/.gitignore` ignores `dist/` and `*.egg-info/` but not `build/`, so setuptools output (~5.3k of the diff's added lines) got committed. `promptopt`'s copy duplicates its source exactly; `dfs_webworker`'s is **stale** — `build/lib/dfs_webworker/forecasting.py` is 1046 lines vs 1229 in source, is missing the aggregation/MASE/CV work, and still has module-level `from dfs_webworker.store import Store` / `import success` (`:24-25`) — the exact top-level `dfs_webworker` import the design forbids because it breaks standalone `.luml` inference. The new `[tool.setuptools.packages.find] include = ["dfs_webworker*"]` in `pyproject.toml` reads as a workaround for `build/` being present. Recommend deleting both `build/` trees and adding `build/` to `wasm/.gitignore`.

## Smaller issues

- **`formatCellValue` can throw on an invalid `Date`** — `frontend/src/components/table-view/index.vue:140-142`. This new `#body` template runs for every cell of every table view (tabular, prompt-fusion, forecasting); `value.toISOString()` throws `RangeError: Invalid time value` for an Invalid Date, which would crash the whole table render. Practically hard to reach (arquero falls back to string when values don't all parse), but it's the one new shared code path that can throw for the non-forecasting tasks — guard with `Number.isNaN(value.getTime())`.

- **Train is enabled without checking the target is numeric** — `frontend/src/hooks/useForecastSetup.ts:42-48`, surfaced at `ForecastingWrapper.vue:85`. `isValid` only validates that the date column parses; the target auto-defaults to the first non-date column (`useForecastSetup.ts:123-126`), so uploading `date, region(string), sales` selects `region`, enables Train, and dispatches a non-numeric target that fails server-side with a toast. A numeric check before enabling Train would fail earlier and clearer.

- **Serialized `"version": 1` is never validated** — `forecasting.py:676`; `from_dict` (`:688-702`) and `_pyfunc.warmup` ignore it. A future v2 spec loaded by v1 code fails with a bare `KeyError` rather than a clear "unsupported pipeline version" error.

- **Minor:** a column that is the date or target has an empty `menuItems` and thus no header menu (`ForecastColumnHeader.vue:115`, `:30`), so reassigning those roles requires a non-obvious multi-step move via another column; and `overlays.knownFutureCols` is threaded through `buildForecastSegments`/`ForecastChart`/the registry card but never used (dead param).

## Missing test coverage

- **The core "runs standalone without `dfs_webworker`" contract is never exercised.** `tests/test_forecasting_serialization.py` runs the fnnx runtime in-process where `dfs_webworker` is importable, so a regression reintroducing a top-level `dfs_webworker` import in `forecasting.py` (exactly what the stale `build/lib` copy has) would pass green. Add a test that drops `dfs_webworker` from `sys.modules`/path before loading a bundle and predicting.
- **`aggregation="sum"` is entirely untested** — no `sum`/`aggregation` reference appears in the Python tests, leaving both the `_aggregate_to_grid` sum path and the `to_dict`/`from_dict` round-trip of `aggregation` uncovered (a regression dropping it from `to_dict` would silently revert a `sum` model to `mean`).
- **`useForecastingTraining` unit tests cover only metric formatting** — `startTraining`/`startPredict`/`saveModel`/`deleteModels` are untested (the Evaluate component tests do exercise the predict flow via a mocked `predict`).
- **`forecasting-setup` tests use only clean ISO/same-day inputs**, so the day-of-month horizon and non-ISO/timezone parsing edge cases above are uncovered.

codie-dfs:
codie started reviewing commit `8e1ba34` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `8e1ba34` (requested by @OKUA1)

I've completed a thorough review of the full PR. Let me compose the review.

## Summary

This PR adds an end-to-end **time-series forecasting** feature: a new Python SARIMAX engine (`dfs_webworker/forecasting.py`) with automatic order selection, rolling-origin CV, JSON-only serialization (no training data or pickled results), and FNNX bundling; plus the full Vue/TS surface (setup wizard, evaluate dashboard, re-forecast with known-future inputs, registry model card) wired through the web worker. Statsmodels/patsy are added to the pyodide worker, and the task is enabled on the home screen.

Overall this is a high-quality, unusually well-tested change — the Python engine has extensive unit/route/serialization tests and the frontend covers the date math, normalizer, chart builder, hooks, and evaluate component including failure paths. I found no clear correctness or security bugs in the core logic. The notes below are low-severity.

## Minor correctness / consistency

**`week` forecast dates are generated from a different anchor than the engine.** `frontend/src/lib/data-processing/forecasting-setup.ts:112` (`lastDate`) takes the *maximum raw* history date, and `forecasting-setup.ts:168` (`forecastDate`) advances the `week`/`day` frequencies by fixed offsets from it. The engine instead anchors on `df[date_col].iloc[-1]`, which after `_aggregate_to_grid(..., "first")` is the *first* date of the last period. For `month`/`quarter`/`year` this is irrelevant (both snap to the period start), and for `day` the period is a single day, but for `week` with more than one history row in the final week the two anchors differ by <7 days. The re-forecast itself still succeeds (the engine validates `future` by week *period*, and +7-day steps preserve the week), but the `supplied` overlay in `frontend/src/components/express-tasks/forecasting/evaluate/index.vue:295` keys frontend future dates against the engine's returned dates, so the "Provided" series silently renders empty in that case. Cosmetic, narrow.

## Smaller problems

- **`_add_future_preview` calls `predict` unguarded during `fit`** (`wasm/packages/dfs_webworker/dfs_webworker/forecasting.py:1089`). The CV folds are defensively wrapped (`_run_fold` returns `None` on exception), but the optional training-time future preview is not — any predict-time error would abort the whole `fit()` rather than just skipping the preview. I checked the `min_history ≤ n` invariant holds given the seasonal gating thresholds (`2·period+1` to enable seasonality, `3·period` to enable seasonal differencing), so the obvious "not enough history" path shouldn't fire; still, wrapping the preview so a non-essential visualization can't sink training would be safer.
- **Unrelated change bundled in:** removing the `time` prop from `UiTraining.vue` (and the `:time="8"` call sites in `TabularWrapper.vue` and `prompt-fusion/step-main/index.vue`) drops the "Estimated loading time" line for the tabular and prompt-fusion flows too. It's complete and consistent, but it's outside the forecasting feature.

## Missing test coverage

- No test exercises the `week`-frequency date alignment with multiple history rows in the final period — this is exactly the case that would surface the anchor mismatch above (all predict/date tests use clean monthly grids).
- The `_add_future_preview` failure path is untested (the preview tests all use `n=48`, well above `min_history`).

Everything else — the exogenous two-stage flow, known-future validation/reordering, MASE/MAPE edge handling, CV fold construction, serialization roundtrip, and the timezone-independent date logic — is correct and thoroughly covered.

OKUA1 (approved):

PR #619, opened 2026-07-09
updated prisma default storage path

OKUA1:


codie-dfs:
codie started reviewing commit `af29639` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `af29639` (requested by @OKUA1)

I now have a complete picture. The source migration is complete and internally consistent — the only remaining old-path references are inside `.codie-review/diff.patch` itself. Config is read-only (never written by the app), which makes the location change silent for existing users.

---

## Summary

This PR consolidates all Prisma storage locations under a single `~/.luml/` root: experiment data at `~/.luml/experiments/` (shared with `luml-sdk`) and Prisma-specific state (main DB, config, upload queue, sessions, custom CLIs) under `~/.luml/prisma/`. Previously these were scattered across `~/.prisma/`, `~/.luml-prisma/`, `~/.local/share/luml-prisma/`, and `~/.config/luml-prisma/`. The rename is applied consistently across code, tests, and docs.

## Correctness

No bugs in the code itself. I verified the paths are internally consistent and the experiment-linking contract stays matched on both sides:

- Writer (`data/guide.md:196`, agent training code) and reader (`api/uploads.py:23` `_link_experiments`) both now point at `~/.luml/experiments`, so linking still works.
- `app.py:86` (`global_dir / "prisma" / "uploads.db"`) resolves to the same `~/.luml/prisma/uploads.db` as the `UploadQueue` default in `upload_queue.py:77`, and `ensure_global_luml_dir()` (`utils.py:26-30`) creates both `experiments` and `prisma` subdirs.
- No source file outside the diff still references any old location.

## Main concern: hard cutover with no migration or fallback

This is a breaking change to on-disk locations with no migration step and no read-fallback to the old paths. On upgrade, any existing installation silently loses access to its data:

- **User config is silently dropped.** `get_config_path()` (`config.py:37-38`) now reads `~/.luml/prisma/config.toml`. `load_config()` only ever *reads* this file (`config.py:53-57`) — it is never written by the app, so it is entirely user-authored. An existing user with a hand-written `~/.config/luml-prisma/config.toml` (custom `cors_origins`, `branch_prefix`, `preserve_patterns`, etc.) will get **silently reverted to defaults** with no error or warning. This is the most user-hostile part of the cutover.
- **Task/repository history is abandoned.** The main DB moves from `~/.local/share/luml-prisma/luml-prisma.db` to `~/.luml/prisma/luml-prisma.db` (`config.py:32,43`); the app will initialize a fresh empty DB at the new path and prior history is no longer read.
- **In-flight uploads are orphaned.** Any `PendingUpload` rows in the old `~/.prisma/uploads.db` are never processed.
- **Custom coding CLIs disappear.** `_custom_agents_path()` (`services/agents.py:72`) moves `coding-clis.json` from `~/.luml-prisma/` to `~/.luml/prisma/`; a user's registered custom agents will no longer be found.
- **Prior experiments vanish from `luml-inspect`**, since the default DB path moved.

If this tool is pre-release / internal and no installed users need to be preserved, this is fine — but it should be a deliberate, documented decision. At minimum, consider (a) a one-time migration that moves the old directories if present, or (b) falling back to the old path when the new one is absent, or (c) logging a warning when an old config/DB exists but the new location is empty. The silent config fallback in particular is worth guarding.

## Test coverage

- `test_prisma_dir.py` is updated for `ensure_global_luml_dir()` (now asserts both `experiments` and `prisma` subdirs) and the guide.md content check — good.
- The relocated paths in `config.py` (`get_data_dir` / `get_config_path`) have **no direct test** asserting the new location; `test_agents.py` monkeypatches `_custom_agents_path` so it doesn't pin the new path either. These are trivial one-liners, so low priority, but a small assertion that `get_config_path()`/`get_data_dir()` resolve under `~/.luml/prisma` would lock in the intended layout and catch accidental regressions.
- No test covers migration/fallback behavior — only relevant if you decide to add one per the concern above.

Overall the mechanical rename is clean and consistent; the only thing I'd want confirmed before merge is that abandoning existing on-disk state (especially user config) is intentional.

PR #625, opened 2026-07-19
feat: change evals view

limanski-valera:


codie-dfs:
codie started reviewing commit `e7c2c85` (requested by @limanski-valera).

codie-dfs (commented):
### codie review of `e7c2c85` (requested by @limanski-valera)

I have enough context to write the review. Let me verify one detail about the race-condition concern — whether tabs stay interactive during loading (they're outside the `showLoader` gate in `TracesDialog.vue`).

The tabs (`EvalsTabs`) are rendered outside the `showLoader` conditional, so they remain clickable while a fetch is in flight — confirming the race is reachable. I now have everything needed.

## Summary

This PR reworks how evals appear in the traces table: instead of rendering every eval id as a separate button, the cell now shows only the first eval plus a `+N` counter, and the `TracesDialog` gains a tab bar (`EvalsTabs`) to switch between the row's evals, backed by new `evalsIdsList` / `currentEvalId` / `currentEvalDataLoading` state in the evals store and a `useMinLoading` hook to avoid skeleton flicker. The logic is coherent and the tab-selection sync (`currentEvalId` ← `EvalsInfo.id`, matched against `evalsIdsList`) is internally consistent since `getEvalById` returns an eval whose `id` equals the requested id.

## Correctness

**`v-else` branch dereferences `evals` unconditionally — a robustness regression** (`extras/js/packages/experiments/src/components/traces/TracesTable.vue:257-267`)

The `v-if` still guards with optional chaining (`slotProps.data.evals?.length === 0`), implying `evals` may be nullish, but the new `v-else` branch reads `slotProps.data.evals[0]` and `slotProps.data.evals.length` with no guard. If `evals` is ever `null`/`undefined`, `evals?.length === 0` is `false`, so the else branch runs and `undefined[0]` / `undefined.length` throws a render error. The previous code (`v-for="item in slotProps.data.evals"`) degraded gracefully to rendering nothing in that case. The `Trace` type declares `evals: string[]` and both providers currently return an array, so this is latent rather than guaranteed — but the change is strictly less defensive than what it replaced. Either drop the now-misleading `?.` on the sibling `v-if`, or guard the else branch (e.g. `v-else-if="slotProps.data.evals?.length"`).

**Out-of-order responses on rapid tab switching** (`extras/js/packages/experiments/src/components/evals/traces/TracesDialog.vue:64-76`)

`onCurrentEvalIdUpdate` has no request sequencing, and the tabs stay interactive during loading (they're rendered outside the `showLoader` gate). If a user clicks tab B then tab C and B's `getEvalById` resolves after C's, the final `setCurrentEvalData` call wins with B's data while C's tab shows as active — the displayed eval won't match the selected tab. Consider tracking the latest requested `evalId` and ignoring stale responses (or aborting in-flight requests). Edge case, but reachable given the 500 ms minimum loader.

## Minor

- **Dialog close doesn't clear the new state** (`extras/js/packages/experiments/src/store/evals/index.ts:147-148` vs the close path). The dialog is closed via `@update:visible="evalsStore.resetCurrentEvalData"` (`ExperimentSnapshot.vue:37`), which only nulls `currentEvalData`; `evalsIdsList` and `currentEvalDataLoading` are cleared only by the full `reset()`. It's harmless today because `showEval` overwrites `evalsIdsList` on the next open, but the reset paths are now inconsistent — worth aligning so `resetCurrentEvalData` (or the close handler) also resets the companion state.
- **`+N` counter looks clickable but isn't** (`TracesTable.vue:265-267`). `.evals-count` uses `--p-text-link-color` and sits next to a real link, but it's a plain `<span>`; the only way to reach evals 2..N is to open eval 1 and use the tabs. If that's intended, consider not styling it as a link to avoid implying an action.

## Missing test coverage

The package has an established `__tests__` convention (e.g. `src/store/evals/__tests__`, `src/providers/__tests__`, component tests), but this PR adds untested behavior:

- `useMinLoading` — the minimum-duration hold, rapid toggle (`stop()` cancelling a pending hide), and the `startedAt`-based remaining-time computation are all logic worth a deterministic test (fake timers).
- New store surface — `setEvalsIdsList` / `resetEvalsIdsList`, `currentEvalId` derivation from `selectedEval`, and that `reset()` now clears `evalsIdsList` / `currentEvalDataLoading`.
- `showEval(evalId, evalsIds)` now setting `evalsIdsList`, and the single-vs-multi eval rendering branch in the table cell (including the empty/`-` case).

codie-dfs:
codie started reviewing commit `86a069e` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `86a069e` (requested by @OKUA1)

Review could not be completed because repository reads failed before commands could start: the read-only sandbox could not initialize its filesystem namespace. No findings are reported because `.codie-review/diff.patch` and its surrounding code could not be inspected; please rerun the review in a functioning checkout environment.

codie-dfs:
codie started reviewing commit `86a069e` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `86a069e` (requested by @OKUA1)

This PR groups trace eval IDs, adds tabbed eval switching in the trace dialog, and introduces a minimum-duration skeleton loader. The asynchronous switching and responsive loading layout need correction before merge.

### Correctness bugs

- **High — stale requests can overwrite newer selections or reopen a closed dialog** (`extras/js/packages/experiments/src/components/evals/traces/TracesDialog.vue:69`). Tab requests are neither cancelled nor sequenced, and every response unconditionally updates shared store state. Clicking B then C can display B if B resolves last; closing the dialog while a request is pending allows its response to repopulate `currentEvalData`, reopening the dialog. Track a request generation or use cancellation, invalidate it on close/unmount, and only commit data and clear loading for the latest active request.

### Smaller problems

- **Responsive layout regression** (`extras/js/packages/experiments/src/components/evals/traces/TracesDialog.vue:94`). `TracesInputs` shrinks to 300px below 992px, but its loading replacement remains 380px, causing an 80px layout shift and severely squeezing the main content on narrow screens. Share the same responsive width rules between both panes.

- Newly added functions omit the project-required explicit return types, including `useMinLoading` (`extras/js/packages/experiments/src/hooks/useMinLoading.ts:4`), `onCurrentEvalIdUpdate` (`extras/js/packages/experiments/src/components/evals/traces/TracesDialog.vue:64`), and `onValueUpdate` (`extras/js/packages/experiments/src/components/evals/traces/EvalsTabs.vue:29`). Add `Ref<boolean>`, `Promise<void>`, and `void` annotations respectively, along with `void` on the new store setters.

### Missing test coverage

No tests were added. Add deterministic tests using deferred promises for out-of-order tab responses and closing during a request, fake-timer coverage for `useMinLoading`, and component coverage for empty, single, and multiple eval-ID rows.

K-Krut (approved):

PR #631, opened 2026-08-24
fix: model containers survive a restart after their artifact URL expires

K-Krut:
## What broke

A deployment's artifact URL is presigned, baked into the model container's environment at creation, and valid for 12 hours. The container lives for weeks. Docker will not let an environment change afterwards, so **any model container restarted more than 12 hours after its deploy died** — it tried to download its model from a dead link, got a 403, exhausted its three restart attempts and stayed `Exited(1)` for good.

On the demo stand that was every deployment at once. The only way back was recreating the deployment on the Platform, since a `DEPLOY` task is only enqueued on creation — there is no redeploy.

Two things that would have absorbed this had been commented out since the satellite first landed: the model cache volume, and the cache-hit branch that uses it. The cleanup path still mounted that volume to delete from it, so undeploy was clearing a cache nobody wrote to.

## What changed

**The container is no longer given a download link.** It carries a token and asks the Agent for a URL at the moment it downloads (`GET /satellites/deployments/{id}/artifact`). A URL that lives seconds cannot expire under a container that lives for weeks, so the failure cannot happen by construction. This is what the spec always described — *"Satellite obtains a short-lived artifact download URL from Platform"* — the URL had simply been handed out early.

Access is proved with a per-deployment token derived from the Satellite credential rather than stored: `HMAC(SATELLITE_TOKEN, deployment_id)`. Nothing to persist means nothing to lose across a restart of either side, and the token binds to one deployment, so a container can only ever reach its own artifact.

**The cache is back on**, keyed by `artifact_id` (it used to be an md5 of the presigned URL's path). The key is pinned into the container's environment at creation, so a model that is already unpacked is found *before* anything is asked of anyone — a container comes back up even while the Agent or Platform is down. Extraction stages into a unique directory and moves into place only once complete, so a crash cannot leave a half-unpacked model that the next start reads as a hit.

**Stopped containers are recreated on startup reconciliation**, with a URL signed at that moment. The rule is deliberately narrow:

- container exists but is stopped → recreate it
- no container at all → leave it, report `not_responding`

The second case looks tempting (the deployment is active, bring it up) but that is a deploy from nothing, not a recovery, and the Platform owns that decision. An earlier version of this change did create containers there and revived 19 long-abandoned deployments at once, each downloading a model and building a conda environment until the machine ran out of disk. `not_responding` deployments are in scope, because that is where a failed reconciliation parks them — without it the first failure would be permanent.

Along the way: `SATELLITE_AGENT_URL` had never pointed anywhere. The hostname had no network alias behind it and the port came from `MODEL_SERVER_PORT`. Nothing noticed because nothing read the variable — and this change depends on it.

## Verified on a live stand

Created a real deployment through the Platform API and watched it through:

- container is created with `MODEL_ARTIFACT_TOKEN` + `MODEL_ARTIFACT_ID` and **no URL**
- it calls the Agent, gets a link signed in the same second, downloads and unpacks
- a second deployment of the same artifact starts with **0 calls to the Agent and 4 cache hits**, reusing the entry the first one left
- endpoint refuses a missing token (403), another deployment's token (403), and a deployment this Satellite does not host (404)

**The race was reproduced for real**, not simulated: two containers of one artifact launched simultaneously on a cold cache both finished unpacking in the same millisecond. One won the rename, the other adopted its copy and carried on. No `.partial` left behind, cache entry intact. On the previous code that same interleaving would have killed the loser with `ENOTEMPTY`.

Unrelated finding, not addressed here: `model_server/Dockerfile` pulls micromamba from `micro.mamba.pm/api/micromamba/linux-64/...` while the image builds `arm64`, so on Apple Silicon the container dies at conda setup with a rosetta error — after the artifact path has already succeeded.

## Deploying this

**The Agent and the model image must be released together.** The environment contract between them changed: `MODEL_ARTIFACT_URL` is gone, `MODEL_ARTIFACT_TOKEN` and `MODEL_ARTIFACT_ID` take its place.

Existing containers keep running until recreated. Stale cache entries under the old md5 key are orphaned — a few hundred KB on the stand, worth a one-off sweep on a real Satellite.

## Tests

186 passing, ~35 new: the artifact endpoint and token binding, the model server's resolution and retry, cache hits and the concurrent-unpack race, and the reconciliation rules including the one that must *not* resurrect a container that does not exist.

`ARTIFACT_URL_PLAN.md` carries the design and the rollout order it was built in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0172UBhZDaVE3RMrmbsw5jhB


codie-dfs:
codie started reviewing commit `46bd14e` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `46bd14e` (requested by @OKUA1)

The PR replaces deployment-time presigned URLs with an HMAC-authenticated Agent endpoint, introduces an artifact-ID cache, and relaunches stopped containers during startup reconciliation. It adds unit coverage for endpoint authorization, cache behavior, download retry, and several recovery paths.

## Correctness bugs

- **High — Existing running deployments are never migrated to the new protocol.** `satellite/agent/handlers/model_server_handler.py:244` only recreates containers already stopped; healthy legacy `sat-*` containers retain their old image and expiring `MODEL_ARTIFACT_URL`. These dynamic containers are not Compose-managed, so rebuilding the model image does not update them. When one later restarts, it still fails on the expired URL, and reconciliation only runs at Agent startup. Add a launcher/protocol-version label and perform a drained replacement of legacy containers during rollout.

- **High — Interrupted recovery permanently strands deployments in `pending`.** `satellite/agent/handlers/model_server_handler.py:142` changes the status to `pending`, while `satellite/agent/handlers/model_server_handler.py:211` deliberately excludes all pending deployments from subsequent reconciliation. If the Agent restarts or `_settle_recovered` raises during the potentially 30-minute health-check window, no deploy task exists to finish that state and future starts ignore the deployment. Keep recovery in a reconcilable status, persist a distinct recovery marker, or only exclude pending deployments that actually have an in-flight deploy task.

- **High — The persistent cache has no active cleanup path.** `satellite/agent/clients/docker_client.py:65` mounts a named volume into every model container, but the sole cleanup call remains commented out at `satellite/agent/tasks/undeploy.py:60`. Consequently every extracted artifact—and any `.partial` directory left by a hard process termination—survives undeploy indefinitely, eventually exhausting disk. Invoke cleanup after the final deployment using an artifact is removed and reclaim orphaned staging directories.

## Security issues

- **High — All model containers can read and modify every cached artifact.** `satellite/agent/clients/docker_client.py:65` mounts the same volume read-write into every deployment, and `satellite/model_server/handlers/model_handler.py:143` trusts any non-empty cache directory. Code running in one model container can therefore enumerate other artifacts or poison another deployment’s cached model for its next restart, bypassing the per-deployment artifact-token boundary. Isolate cache mounts per deployment/artifact, or populate an Agent-controlled cache that model runtimes receive read-only with integrity validation.

## Missing test coverage

- Add an upgrade test for a healthy legacy container followed by a later restart.
- Test an Agent restart after recovery writes `pending` but before it writes `active`.
- Cover final-undeploy cleanup, orphaned `.partial` directories, and cross-container cache isolation.
- The 403 retry test should return two distinct URLs and assert two Agent calls; currently it would also pass if the stale URL were simply retried.
- The Docker tests mock the engine, so no test verifies the Compose alias is actually resolvable from a dynamically created model container.

OKUA1 (approved):

codie-dfs:
codie started reviewing commit `9f71544` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `9f71544` (requested by @K-Krut)

This PR moves artifact URL resolution into the Agent, adds per-artifact cache volumes and atomic extraction, and migrates/relaunches containers during startup reconciliation. The core flow is coherent, but recovery currently risks corrupting Platform state or permanently stranding deployments.

## Correctness bugs

- **High — successful recovery clears deployment metadata.** `satellite/agent/handlers/model_server_handler.py:158` and `satellite/agent/handlers/model_server_handler.py:192` use the general deployment PATCH for recovery status changes. `backend/luml/handlers/deployments.py:197` converts omitted fields into explicit `None`, and `backend/luml/repositories/deployments.py:99` persists them, erasing `inference_url`, schemas, and tags. Consequently, migrated deployments can be marked active without routing metadata. Make the backend PATCH preserve unset fields, or restore the same metadata as the normal deploy finalization.

- **High — a failed replacement can permanently strand a deployment.** `satellite/agent/handlers/model_server_handler.py:141` calls the destructive `create_or_replace` before persisting the recovery marker. If replacement removes the old container and subsequent creation fails, the catch records a generic failure; the next reconciliation reaches `satellite/agent/handlers/model_server_handler.py:253` and deliberately refuses to recreate the now-missing container. Persist recovery ownership before replacement and permit missing-container retries only for that marker.

## Security issues

- **Medium — deployment environment variables can redirect the artifact credential.** User-provided values overwrite internal values at `satellite/agent/handlers/container_launcher.py:67` and `satellite/agent/clients/docker_client.py:67`. A deployment can override `SATELLITE_AGENT_URL`, causing the genuine per-deployment token to be sent to an arbitrary host, or override the token/identity/cache-key variables and break isolation. Reject reserved keys or merge internal runtime variables last.

## Smaller problems

- `satellite/agent/agent_api.py:97` uses `Header(default=None)` as a function-call default without the configured `B008` exemption, so the Ruff CI job should fail. Prefer `Annotated[str | None, Header()] = None`.

## Missing test coverage

- The reconciliation tests mock Platform PATCH responses, so they do not exercise the real partial-update semantics or verify preservation of routing metadata.
- Add a two-pass regression where replacement removes the old container, creation fails, and the next reconciliation safely retries.
- `satellite/tests/integration/test_smoke_agent_alias.py:10` verifies DNS only. The critical cold-cache path still lacks an end-to-end test covering container → Agent → Platform URL → download → restart from cache.

codie-dfs:
codie started reviewing commit `2dc90b4` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `2dc90b4` (requested by @K-Krut)

This PR moves artifact URL resolution to download time, adds authenticated Agent callbacks and per-artifact caches, and relaunches legacy/stopped containers. It also fixes partial deployment updates and reorganizes Satellite tests.

### Correctness bugs

- **High — recovery can cancel an in-progress deletion.** `satellite/agent/handlers/model_server_handler.py:225` unconditionally promotes a recovered deployment using a stale snapshot. If deletion is requested during the potentially long health wait, this overwrites `deletion_pending` with `active`; the undeploy task then removes the container, but the backend rejects deletion because the status is no longer pending, leaving `deletion_failed`. The marker write at `satellite/agent/handlers/model_server_handler.py:146` has the same race. Make recovery transitions conditional on the current status/marker and never transition out of deletion states.

- **Medium — `health_check_timeout` is treated as an attempt count, not a time limit.** At `satellite/agent/handlers/model_server_handler.py:212`, every iteration can spend five seconds in the request plus one second sleeping. The default “1800-second” recovery can therefore take roughly three hours against a hanging endpoint. Use a monotonic deadline or `asyncio.timeout`.

### Security issues

- **Medium — the artifact token can be sent through an environment-configured proxy.** `satellite/model_server/clients/agent_client.py:56` uses the default `httpx` environment proxy handling, while deployment environment variables can include `HTTP_PROXY`/`ALL_PROXY`. The internal Agent request may consequently fail or disclose the long-lived artifact token to that proxy, despite locking down `SATELLITE_AGENT_URL`. Disable environment proxies for this internal request or explicitly bypass them for the Agent host.

### Missing test coverage

- Add tests for deletion racing with recovery, elapsed timeout behavior with a hanging health endpoint, and proxy variables during Agent resolution.
- `satellite/tests/integration/test_smoke_agent_alias.py:48` verifies only DNS resolution; an integration test should exercise the actual authenticated artifact callback and download from a launched model container.

codie-dfs:
codie started reviewing commit `2f82cd2` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `2f82cd2` (requested by @K-Krut)

This PR replaces container-baked presigned URLs with an HMAC-authenticated Agent endpoint, per-artifact cache volumes, and atomic extraction. It also expands startup reconciliation and makes worker deployment updates sparse and deletion-aware.

### Correctness bugs

- **High — Deletion can leave a replacement container running.** `satellite/agent/handlers/model_server_handler.py:237` returns on the backend’s deletion-related `400`, but removes the recreated container only for `404`. If undeploy removes the old container before recovery creates its replacement, that replacement can survive deletion; the orphan scan also excludes IDs from the initial snapshot at `satellite/agent/handlers/model_server_handler.py:325`. Stop the replacement on a deletion refusal or compare every managed container against a fresh deployment list.

- **Medium — Last-undeploy cache cleanup races with container replacement.** `satellite/agent/clients/docker_client.py:251` checks container references and then deletes the volume without locking or considering deployment records. During `create_or_replace`’s delete-then-create gap, a live deployment has no container reference, so another undeploy can delete its warm cache. Use the same live-artifact protection as `cleanup_stale_staging`, or serialize launch/cleanup per artifact.

- **Medium — The model client can time out before valid artifact resolution completes.** `satellite/model_server/clients/agent_client.py:16` allows 30 seconds, while the resolver performs two sequential Platform calls at `satellite/agent/main.py:55` and `satellite/agent/main.py:62`, each with a 30-second timeout. Two individually successful slow calls can therefore crash-loop the model container. Use one scoped Platform request or set an end-to-end timeout exceeding the resolver’s total budget.

### Smaller problems

- `satellite/agent/handlers/model_server_handler.py:448` treats every `4xx` as a terminal refusal. Transient responses such as `408` or `429` leave a healthy container unregistered until the Agent restarts. Handle only deletion-specific responses as terminal and retry rate-limit/time-out responses.

### Missing test coverage

- The deletion-race test should verify that a replacement container is removed after a `400` and when the record disappears after the initial reconciliation snapshot.
- Cache tests do not cover a live deployment temporarily having no container during replacement.
- `satellite/tests/integration/test_smoke_agent_alias.py:23` skips unless Compose is already running, so the existing CI job does not exercise the new network alias or an end-to-end artifact download after restart.

codie-dfs:
codie started reviewing commit `f0b6d7b` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `f0b6d7b` (requested by @K-Krut)

This PR moves artifact URL resolution to the Satellite Agent, adds per-artifact caches and container recovery, and hardens deployment status updates against deletion races. Several upgrade and recovery paths can still leave deployments unavailable or bypass the new artifact contract.

## Correctness bugs

- **High — Existing Satellites will not resolve the new Agent hostname.** `satellite/agent/clients/docker_client.py:35` now requires the `satellite-agent` alias, but that alias is added only to `satellite/docker-compose.yml:44`. Pulling a new Agent image does not update an existing installation’s Compose file, where only the `agent` service name resolves. Cache misses and proactive legacy-container replacement will therefore fail after an image-only upgrade. Use the existing service DNS name, make the address configurable, or enforce a Compose migration before enabling this path.

- **High — Transient Agent or Platform failures can permanently strand a container.** `satellite/model_server/handlers/model_handler.py:141` performs artifact resolution once; transport errors and 5xx responses escape initialization. Docker permits only three restarts at `satellite/agent/clients/docker_client.py:84`, and reconciliation runs only at Agent startup, so several quick 502 responses can leave the container exited after the Platform recovers. Retry transient resolution failures with backoff, use an unlimited restart policy, or periodically reconcile stopped containers.

- **Medium — Orphan cleanup leaves deleted deployments registered locally.** `satellite/agent/handlers/model_server_handler.py:346` uses a fresh Platform listing and removes containers missing from it, but never removes those IDs from `self.deployments`. An active deployment registered from the opening snapshot can be deleted before this second listing; subsequent compute requests still accept it, and if the suppressed Docker removal fails, it can continue serving after deletion. Remove the local registration regardless of Docker cleanup success.

## Security issues

- **High — Deployment environment variables can bypass artifact authorization.** `satellite/agent/handlers/container_launcher.py:53` preserves a deployment-supplied `MODEL_ARTIFACT_URL`, and `satellite/model_server/handlers/model_handler.py:118` prefers it even when the protocol-2 token is configured. This lets the container download an arbitrary archive instead of the Platform-authorized artifact. Strip this variable in protocol-2 launches and only honor the legacy fallback when no artifact token is present.

## Missing test coverage

Coverage is missing for upgrades using a pre-change Compose network, transient resolver failures followed by recovery, deletion of an active deployment between the two reconciliation snapshots, and protocol-2 deployments supplying the legacy artifact URL through either environment-variable source.

PR #634, opened 2026-08-28
Feature/lm 649 update registry design

limanski-valera:


codie-dfs:
codie started reviewing commit `030e5af` (requested by @limanski-valera).

codie-dfs (commented):
### codie review of `030e5af` (requested by @limanski-valera)

This PR introduces shared rich cards and selectors, adds collection tag filtering, and expands collection/track metadata and validation. The tag-filter implementation currently has blocking rendering and state issues, while the shared tag row regresses collections whose tags are nullable.

## Correctness bugs

- **High — The tag filter panel is hidden.** `frontend/src/components/ui/UiRichTagsSelect.vue:41` spreads `COLLECTION_TYPE_SELECT_PT`, which sets the MultiSelect header to `display: none` at `frontend/src/components/orbits/tabs/registry/collection.const.ts:21`. Because `UiRadioSelect` is rendered in that header and `listContainer` is also hidden, opening the Tags control produces no visible options. Override/remove the inherited header rule for this component.

- **High — Collections with `null` tags crash card setup.** `frontend/src/components/orbits/tabs/registry/CollectionCard.vue:11` passes the API value directly, while `frontend/src/components/ui/UiTagsRow.vue:33` immediately accesses `props.tags.length`. The backend permits nullable collection tags, including when collections are created without tags. Normalize with `data.tags ?? []` and correct the frontend interface to reflect nullability.

- **Medium — “show all” selects nothing in the tag picker.** At `frontend/src/components/ui/UiRadioSelect.vue:94`, the checked branch calls `fillSelectedOptions(options, [])`; that helper interprets an empty selection using `selectedByDefault`, which the tag picker sets to `false`. Set every option to selected explicitly instead of reusing the initialization sentinel.

- **Medium — Selector state does not react to updated inputs.** `frontend/src/components/ui/UiRadioSelect.vue:65` snapshots `options` and `selectedOptions` only during setup. If the tag panel opens before the asynchronous tag request finishes, it remains empty after the response; external selection changes can likewise leave stale switches. Watch the relevant props or rebuild state when the panel opens.

- **Medium — Tag options are fetched only once.** `frontend/src/components/orbits/tabs/registry/CollectionsToolbar.vue:105` loads tags only in `onBeforeMount`. Vue Router reuses this view when switching between orbit IDs, and create/update/delete operations also never refresh `collectionsTags`, so the selector can show another orbit’s tags or omit newly added ones. Refresh on orbit changes and collection tag mutations.

## Smaller problems

- **Low — New icon-only controls lack accessible names.** The clear buttons at `frontend/src/components/orbits/tabs/registry/CollectionsToolbar.vue:16` and the description button at `frontend/src/components/orbits/tabs/registry/collection/CollectionHeader.vue:17` expose no contextual label. Add appropriate `aria-label` values.

- **Low — Added functions omit the required explicit return types.** Examples include `frontend/src/lib/api/orbit-collections/index.ts:69`, `frontend/src/components/ui/UiRadioSelect.vue:80`, and `frontend/src/components/ui/UiTagsRow.vue:40`. Add `Promise<string[]>`, `Column[]`, `void`, and corresponding annotations throughout the new code.

## Missing test coverage

No tests were added for tag filtering or the new shared components. Add component coverage for nullable tags, asynchronous option updates, and `selectedByDefault=false` “show all” behavior, plus integration coverage confirming repeated `tags` query parameters and orbit switching.

The existing collection integration tests also need updating: selectors at `frontend/tests/integration/collections.spec.ts:168` and `frontend/tests/integration/collections.spec.ts:262` target markup/text removed by this PR, while the broad mock at `frontend/tests/integration/collections.spec.ts:42` also matches the new `/collections/tags` endpoint but returns the wrong response shape.

K-Krut (approved):

PR #636, opened 2026-08-31
fix/satellite e2e findings

K-Krut:


codie-dfs:
codie started reviewing commit `e71ffb9` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `e71ffb9` (requested by @K-Krut)

This PR improves recovery polling, Greptime table initialization, model-environment setup, and SDK dependency discovery. Three issues remain in the recovery and environment-building changes.

## Correctness bugs

- **High:** `satellite/model_server/handlers/model_handler.py:275` treats every `opentelemetry-*` distribution as sharing one version line. Packages such as `opentelemetry-semantic-conventions` and `opentelemetry-instrumentation-*` use 0.x versions alongside 1.x API/SDK packages; if one appears first, the worker dependencies receive invalid 0.x API/SDK/exporter pins and environment creation fails. Restrict version sources to known compatible core distributions instead of matching the entire prefix.

- **Medium:** `satellite/agent/handlers/model_server_handler.py:324` assumes `ContainerNotFoundError` definitively means removal, but `satellite/agent/clients/docker_client.py:141` converts every Docker error from `containers.get()`—including daemon 5xx responses—into that exception. A transient Docker failure therefore ends the one-time startup reconciliation and leaves a healthy replacement unregistered until the Agent restarts. Only map an actual 404 to `ContainerNotFoundError`; retry or tolerate other Docker errors.

## Smaller problems

- `satellite/model_server/handlers/model_handler.py:300` runs environment removal with `check=False` and then always logs that removal succeeded. A lock, permission failure, or other nonzero exit leaves the poisoned environment in place while hiding the cleanup failure. Check the return code and log failure without masking the original build exception.

## Missing test coverage

- Cover a beta-version OpenTelemetry package preceding a stable API/SDK pin.
- Exercise a non-404 Docker error during the new presence check.
- Cover a nonzero micromamba removal result.

codie-dfs:
codie started reviewing commit `66bfea6` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `66bfea6` (requested by @K-Krut)

This PR hardens Docker recovery, refreshes local deployment metadata after Platform updates, handles not-yet-created Greptime tables, and improves model dependency detection/environment setup. The OpenTelemetry compatibility logic and Docker exception contract introduce deployment failures that should be addressed before merging.

### Correctness bugs

- **High — beta OpenTelemetry packages still constrain the core version** (`satellite/model_server/handlers/model_handler.py:287`). `_OTEL_CORE` ignores `opentelemetry-semantic-conventions` and instrumentation packages, so a model containing only `opentelemetry-semantic-conventions==0.63b1` receives worker packages pinned to the image’s 1.43 line. `opentelemetry-sdk==1.43.0` requires its matching 0.64 semantic-conventions release, making that environment unsatisfiable. Derive the compatible core release from these 0.x packages or otherwise let the resolver choose compatible worker versions.

- **High — exhausted Docker inspection errors strand deployments in `pending`** (`satellite/agent/clients/docker_client.py:156`, `satellite/agent/tasks/deploy.py:146`). The method now raises non-404 daemon errors after retrying, but the deploy health loop catches only `ContainerNotFoundError` and `ContainerNotRunningError`. The controller then marks only the task failed; the deployment remains pending, its container remains running, and startup reconciliation ignores pending records. The deploy path should explicitly handle the new “Docker state unknown” outcome, either continuing within the health deadline or consistently failing the deployment and cleaning up.

- **Medium — the final container inspection bypasses the new retry and 404 classification** (`satellite/agent/clients/docker_client.py:164`). `container.show()` performs the status/label read outside the retry block. A 5xx there is not retried, and a 404 caused by removal between `get()` and `show()` escapes as a raw `DockerError` instead of `ContainerNotFoundError`. Include the status read in the retry/classification block.

### Missing test coverage

- Add a resolver-level environment test for artifacts containing only an older semantic-conventions or instrumentation pin; the current unit test merely asserts the generated list and therefore accepts an unsatisfiable set.
- Add a deploy-task regression test for a Docker 5xx/timeout after the container has started, verifying both task and deployment state.
- Add inspection tests where `container.show()` raises 404 and 5xx.

olha-poliuliakkh (approved):

codie-dfs:
codie started reviewing commit `153710a` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `153710a` (requested by @K-Krut)

This PR hardens Docker reconciliation and recovery, refreshes deployment metadata, improves Greptime retention handling, and adjusts model dependency resolution. Several edge cases can still leave deployments unavailable or metadata inconsistent.

## Correctness bugs

1. **High — Transport failures bypass Docker retries.** At `satellite/agent/clients/docker_client.py:155`, only `DockerError` enters the retry loop. A transport-level `TimeoutError` or connection exception from `get()`/`show()` escapes immediately and is skipped by `satellite/agent/handlers/model_server_handler.py:453`; because startup reconciliation runs only once, a healthy deployment can remain unregistered until the Agent restarts. Retry known transport exceptions while preserving the special 404 translation, or schedule another reconciliation.

2. **Medium — Conditional requirements can select the wrong OpenTelemetry version.** `satellite/model_server/handlers/model_handler.py:289` collapses requirements by package name, while `_pinned_version()` at line 267 ignores environment markers and the dependency’s `condition`. With mutually exclusive pins for different Python versions, whichever appears last determines the worker packages, potentially conflicting with the pin active in the target conda environment. Parse and evaluate markers against the target environment configuration, or leave worker dependencies unpinned when constraints are conditional.

3. **Medium — A failed finalization leaves cached metadata as `active`.** `satellite/agent/tasks/deploy.py:190` updates local metadata before reporting the task as done. If that task-status request fails, the exception path PATCHes the deployment to `failed` at line 202 but never applies that returned record locally, so the dashboard continues showing `active`. Capture and apply the fallback PATCH response as well, ensuring the cache reflects the final successful Platform update.

## Smaller problems

- `satellite/agent/monitoring/greptime.py:275` queues every missing TTL target for repeated retries, including the five legacy metric tables at line 265. Current telemetry no longer creates those tables, so a fresh installation will issue five failed `ALTER TABLE` requests every retry interval indefinitely. Treat legacy-table retention as a one-time best-effort operation, or otherwise stop retrying tables that cannot appear.

## Missing test coverage

Add regression tests for:

- A transport `TimeoutError` or connection failure from Docker inspection.
- Mutually exclusive OpenTelemetry pins evaluated against the target Python version.
- Failure of the post-activation task-status update and the resulting local metadata.
- Permanently absent legacy metric tables not remaining in the TTL retry set.

codie-dfs:
codie started reviewing commit `a6cf5c2` (requested by @K-Krut).

OKUA1 (approved):

codie-dfs (commented):
### codie review of `a6cf5c2` (requested by @K-Krut)

This PR adds resilient Docker reconciliation, deployment metadata refreshes, Greptime table retries, and improved dependency discovery. Three correctness gaps remain in conditional dependency handling and recovery retries.

### Correctness bugs

- **High — Conditional requirements can suppress mandatory worker packages.** At `satellite/model_server/handlers/model_handler.py:304`, any normalized package name is treated as installed even when all its requirements have environment markers. For example, a Python 3.12 environment containing `Uvicorn[standard]; python_version < "3.11"` gets no unconditional `uvicorn`, so `conda_worker.py` fails at import. Track unconditional requirements separately and add an unconditional, unpinned worker requirement when only conditional occurrences exist.

- **Medium — The second reconciliation uses a stale Platform record.** At `satellite/agent/handlers/model_server_handler.py:571`, records are retried after a 15-second delay without refetching their status or existence. If an active record is deleted during that delay, the stale active snapshot can be registered locally; the later orphan sweep removes its container but leaves the local deployment entry behind. Refetch each unanswered record before retrying and skip records that disappeared or left a serving status.

- **Medium — Docker inspection can overrun the recovery deadline before health is checked.** `satellite/agent/handlers/model_server_handler.py:259` awaits an inspection that itself performs three potentially blocking attempts at `satellite/agent/clients/docker_client.py:152`. An unresponsive daemon can therefore delay registration of an already-healthy container far beyond its configured health-check timeout. Bound inspections by the remaining deadline or otherwise ensure Docker probing cannot block the health check indefinitely.

### Missing test coverage

- Add cases for a worker package whose marker is false for the target Python, a Platform record deleted during the delayed second look, and a slow Docker inspection under a short recovery timeout.

PR #643, opened 2026-09-03
feat/artifact lineage

K-Krut:


codie-dfs:
codie started reviewing commit `ee86916` (requested by @OKUA1).

OKUA1 (approved):

codie-dfs:
codie review failed. Re-request a review from `codie-dfs` to retry.

codie-dfs:
codie started reviewing commit `ee86916` (requested by @K-Krut).

codie-dfs:
codie review failed. Re-request a review from `codie-dfs` to retry.

codie-dfs:
codie started reviewing commit `ee86916` (requested by @K-Krut).

codie-dfs:
codie review failed. Re-request a review from `codie-dfs` to retry.

codie-dfs:
codie started reviewing commit `ee86916` (requested by @K-Krut).

codie-dfs:
codie review failed. Re-request a review from `codie-dfs` to retry.

codie-dfs:
codie started reviewing commit `ee86916` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `ee86916` (requested by @K-Krut)

This PR adds persisted artifact-lineage graphs across the backend, web editor, Python SDK, and lumlflow publishing. It also preserves deleted artifacts as lineage snapshots and records experiment-to-model relationships during upload.

## Correctness bugs

- **High — edits made while a save is in flight can be silently lost.** `frontend/src/stores/lineage/index.ts:284` builds the submitted batch before awaiting the request, but the canvas and keyboard shortcuts remain active. Any subsequent edit is absent from the request; after the response, `replaceCanvasWithoutHistory(snapshot())` clears its history and the reload overwrites it with server state. Repeated Ctrl+S can also submit the same batch concurrently. Add store-level `isSaving` protection and either block mutations/saves or preserve changes made after the submitted snapshot.

- **High — a failed load after changing focal artifacts leaves the previous graph editable.** `frontend/src/stores/lineage/index.ts:151` replaces state only after a successful request, while `frontend/src/pages/collection/artifact/LineageView.vue:126` merely shows a toast on failure. After navigating within an orbit, the old graph can therefore remain under the new artifact route; saving it sends valid old node IDs to the orbit-scoped batch endpoint and can mutate the wrong lineage. Track which artifact the loaded state belongs to and clear or disable mismatched state when the focal-artifact load fails.

- **Medium — artifact deletion can commit and then report failure.** The calls at `backend/luml/handlers/artifacts.py:466` each use separately committed repository sessions. If cleanup at `backend/luml/handlers/artifacts.py:470` fails, the endpoint returns an error even though the artifact is already permanently deleted, and a retry receives a 404. Perform snapshotting, artifact deletion, and cleanup in one shared transaction, or ensure post-commit cleanup failure cannot turn a successful deletion into a failed response.

- **Medium — excluding linked artifacts can prevent access to later pages.** `frontend/src/hooks/useArtifactsList.ts:49` filters only the already-fetched page. If all 20 items on the first page are locked, `ArtifactsList` receives an empty list and does not mount its virtual scroller, so no lazy-load event can fetch the next cursor even when eligible artifacts exist. Apply exclusions server-side or automatically fetch until an eligible item appears or pagination is exhausted.

## Security issue

- **Medium — the 200-node availability bound is bypassed for direct neighbors.** The `level > 0` condition at `backend/luml/repositories/lineage.py:404` exempts the focal node’s first level, after every incident edge has already been loaded. A focal artifact can accumulate arbitrarily many deleted neighbors over time, producing an unbounded query and response despite the UI and SDK promising a 200-artifact cap. Enforce a deterministic query limit or a bounded per-node degree instead of exempting level one.

## Smaller problem

- Position-only updates for artifacts without an existing lineage node are silently ignored at `backend/luml/handlers/lineage.py:189`. This affects the synthetic focal node shown for an empty graph: moving it enables Save, the API succeeds, and reload resets the position. Either create a node for a valid artifact position reference or prevent the UI from treating that position as persistable.

## Missing test coverage

Add coverage for edits and repeated saves during a pending request, a focal-artifact switch followed by a failed graph load, cleanup failure after artifact deletion, an entirely excluded first selector page with eligible later results, and saving the position of an empty graph. The traversal test should also assert the public 200-node maximum for a high-degree focal node rather than expecting the first level to exceed it.

codie-dfs:
codie started reviewing commit `7e67db2` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `7e67db2` (requested by @K-Krut)

This PR adds persisted artifact lineage end to end, including backend graph storage/API, SDK and tracker propagation, artifact publishing integration, and a Vue Flow editor. The main correctness gaps involve destructive operations on incomplete graphs and request/transaction races that can leave unexpected lineage state.

## Correctness bugs

- **Destructive edits operate on only the visible portion of a truncated graph.** The graph may omit connections after 200 nodes, but `isEditable` does not consider `truncated`, and unlink/replace only manipulate edges currently loaded on the canvas (`frontend/src/stores/lineage/index.ts:182`, `frontend/src/stores/lineage/index.ts:266`, `frontend/src/stores/lineage/index.ts:295`). Consequently, unlinking can leave hidden connections to the artifact, while replacing can leave hidden connections attached to the old artifact; either artifact can reappear after reload. Make truncated graphs read-only for topology changes, or implement backend operations that atomically remove or rewire all incident edges.

- **Accepting “Discard changes” while a save is in progress still lets that save commit.** The route guard resolves successfully after calling `discardChanges` (`frontend/src/pages/collection/artifact/LineageView.vue:132`), but `discardChanges` is a no-op while `isSaving` is true (`frontend/src/stores/lineage/index.ts:321`). Navigation therefore proceeds while the outstanding request continues and persists the supposedly discarded edits. The guard should block or await the save rather than offering a discard action that cannot be honored.

- **Lineage is committed before fallible upload initialization completes.** Artifact creation commits `link_inputs` at `backend/luml/handlers/artifacts.py:283`, then obtains the storage client and upload details at `backend/luml/handlers/artifacts.py:296`. If either later operation fails, the request returns an error but the new lineage nodes and edges remain for an artifact whose upload was never initiated. Generate the upload details before committing lineage, or compensate all artifact and graph state on subsequent failures.

## Smaller problems

- **Concurrent deletion of the last live artifacts in a component can defeat orphan cleanup.** Each deletion runs independently (`backend/luml/handlers/artifacts.py:468`), while the cleanup query treats every currently visible non-null artifact as a reachability root (`backend/luml/repositories/lineage.py:288`). Two transactions deleting different final artifacts can each observe the other as live, both skip cleanup, and then commit an entirely orphaned component. Serialize deletion and cleanup per orbit/component, or perform cleanup after the deletion transaction commits.

- **An empty create-lineage request bypasses source artifact validation.** `target_artifact_ids` permits an empty list (`backend/luml/schemas/lineage.py:43`), which produces an empty batch without resolving the path artifact (`backend/luml/handlers/lineage.py:298`). A request for a missing or foreign source artifact therefore returns an empty success instead of `404`. Require at least one target or explicitly validate the source before returning.

## Missing test coverage

- Backend tests should cover successful lineage linking followed by storage initialization failure, concurrent deletion of the final artifacts in one component, and an empty target list against an invalid source artifact.
- Frontend tests should cover navigation during an unresolved save and unlink/replace attempts when `truncated` is true.

olha-poliuliakkh (approved):

PR #645, opened 2026-09-04
docs: spec for artifacts batch delete

K-Krut:


codie-dfs:
codie started reviewing commit `1fcdc18` (requested by @K-Krut).

codie-dfs:
codie review failed. Re-request a review from `codie-dfs` to retry.

codie-dfs:
codie started reviewing commit `0e8aab2` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `0e8aab2` (requested by @K-Krut)

This PR specifies collection-scoped batch artifact deletion across the backend, web client, and Python SDK, including per-item blockers and forced deletion. Several race conditions and recovery paths currently violate the proposed safety and resumability guarantees.

## Correctness bugs

- **High — deleting `pending_upload` artifacts can create orphaned objects.** `SPEC.md:210` makes active uploads eligible, but their presigned PUT or multipart-completion URLs remain valid. The bucket DELETE can return 404, the database row can be removed, and then the upload can finish and recreate the object; deleting an object also does not abort an S3 multipart upload. Block `pending_upload`, or add an upload-cancellation/abort protocol before removing its record.

- **High — checking references again at confirmation is too late.** `SPEC.md:225` accepts deployments or track links appearing after phase 1, after which the client may already have deleted the object. The existing creation paths at `backend/luml/handlers/deployments.py:126` and `backend/luml/handlers/tracks.py:277` do not reject `pending_deletion` artifacts, so the resulting deployment or track can reference an artifact whose file is gone. Reference creation and the phase-1 transition need a shared transactional interlock, and new references should require an `uploaded` artifact.

- **High — a lost confirmation response is not safely resumable.** `SPEC.md:240` assumes a network error means the confirmation did not commit. Because phase 3 uses separate transactions, the server may have deleted some or all records before the response is lost; retrying then returns `not_found` under `SPEC.md:189`, leaving stale frontend rows and causing single-item SDK deletion to raise. Add an idempotency/operation token or reconcile previously accepted IDs so a repeated confirmation can recover committed outcomes.

- **Medium — force is not actually restricted to storage failures.** `SPEC.md:214` permits forced confirmation for artifacts in any status without a preceding attempt, and the SDK/direct REST endpoint exposes this behavior. That can remove a healthy artifact while leaving its object behind, contradicting the narrower guarantee at `SPEC.md:65`. Enforce a recorded `deletion_failed` state or server-issued failure token, or explicitly scope the safety claim to the web UI.

## Smaller problems

- `SPEC.md:263` retains an all-or-nothing active-deployment pre-check, contradicting the partial-success guarantee at `SPEC.md:55`. An eligible artifact selected alongside an active deployment is not processed at all. Either send the selection through the batch classifier or qualify the guarantee.

- `SPEC.md:266` sends an all-`deletion_failed` selection directly to force deletion, so the “Try again” instruction at `SPEC.md:302` is unavailable for a single failed artifact. Provide separate Retry and Force choices, with normal retry remaining the default.

- `SPEC.md:625` derives a different test database name from the one created and dropped by `backend/tests/conftest.py:64`. On a clean setup the suite targets a database it never created; on a pre-provisioned setup it can mutate an unintended database. Use one explicit test database name and assert that it is a test database before migrations or cleanup.

## Missing test coverage

The planned suites should additionally cover:

- An upload or multipart completion finishing after phase 1.
- Concurrent deployment and track creation around the phase-1 transition.
- A phase-3 response lost after all or only some transactions commit.
- Failure of the status update after a bucket deletion error.
- Duplicate IDs split across client chunk boundaries.

codie-dfs:
codie started reviewing commit `404da17` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `404da17` (requested by @K-Krut)

This PR adds a specification and implementation plan for client-orchestrated, three-phase batch artifact deletion across the backend, web UI, and Python SDK. Several concurrency and result-contract issues could otherwise produce broken deployments or inconsistent client behavior.

## Correctness bugs

- **[P1] Make deployment creation atomic with the deletion transition — `SPEC.md:211`.** The proposed status check in `SPEC.md:710` is a handler-level read performed separately from the deployment insert, while phase 1 also checks blockers separately from moving the artifact to `pending_deletion`. A deployment can read `uploaded`, phase 1 can return a delete URL, and then the deployment can commit before the client deletes the object; phase 3 preserves the row because of the foreign key, but the deployment now references a missing object. Both flows need to lock the artifact row and perform their check plus write in the same transaction.

- **[P2] Track only unresolved artifacts after request-level errors — `SPEC.md:255`.** If phase 1 already classified one artifact as blocked and phase 3 then fails for another, the blocked artifact is complete, not “not completed.” Additionally, `SPEC.md:353` says the SDK raises an ordinary status error, which cannot return the previously merged results promised here. Define an interruption payload or exception carrying `deleted`, resolved `failed`, and the genuinely unresolved IDs.

- **[P2] Preserve existing blockers after forcing storage failures — `SPEC.md:338`.** This says to close the dialog whenever the forced subrequest succeeds, but `SPEC.md:553` requires a pre-existing deployment failure to remain visible. Merge the forced result into the original failures, removing only successfully forced IDs, and close only when the combined failure list is empty.

## Security issues

- **[P2] Validate artifact scope before exposing its status — `SPEC.md:710`.** The existing deployment flow looks up an artifact globally and validates its collection/orbit afterward. Checking status “right after the artifact lookup” would let a caller authorized in one orbit distinguish a non-uploaded artifact in another orbit via the status-bearing 409 response. Scope the artifact query to the addressed orbit or perform the collection/orbit validation before the status check.

## Smaller problems

- **[P2] Supply metadata for SDK-created storage failures — `SPEC.md:166`.** Phase 1 URL entries contain only `artifact_id` and `url`, while every failure must contain the artifact name (`SPEC.md:183`). The SDK accepts only IDs, so it cannot populate `name` when it creates a phase-2 `storage_error`. Include the name in each URL entry, explicitly permit it to be null for client-side failures, or specify an additional lookup.

- **[P3] Correct the SDK force condition — `SPEC.md:737`.** This task says the `force=False` keyword skips phases 1 and 2, contrary to the rest of the design where only `force=True` does so. State the condition explicitly to avoid implementing forced deletion as the default.

## Missing test coverage

Make the planned tests explicitly cover:

- Concurrent deployment creation versus phase-1 deletion.
- A non-uploaded artifact outside the addressed orbit without status disclosure.
- A phase-3 request error after other artifacts in the same chunk were already classified.
- Successful force deletion alongside an untouched deployment/track blocker.
- SDK phase-2 failures retaining the specified artifact name and partial results after a later-chunk error.

codie-dfs:
codie started implementing this spec on branch `codie/feat/LM-587-batch-artifact-deletion`.

codie-dfs:
codie finished. PR opened from `codie/feat/LM-587-batch-artifact-deletion` into `feat/LM-587-batch-artifact-deletion`.

PR #646, opened 2026-09-05
fix(prisma): use supported ExperimentTracker APIs in guide

orbie-dfs:
Prisma agents copy tracking code from the packaged guide, but both examples call nonexistent `ExperimentTracker` methods and fail at runtime. Update them to use `start_experiment`, `log_static`, `log_dynamic`, and `end_experiment`, retaining model linking in the artifact example.

Also correct the imports to `luml.experiments.tracker`, matching Prisma's existing code: the current SDK does not export `ExperimentTracker` from its top-level package.

Add a regression test that extracts every `tracker.<method>(` and `exp.<method>(` call from the guide's fenced Python blocks and checks each against the SDK class. It also asserts that calls were found to prevent an empty extraction from silently passing.

Validation:
- Confirmed the regression test fails against the original guide, identifying all four nonexistent methods.
- `uv run --frozen pytest tests/test_prisma_dir.py -q`: 15 passed; also passed with `PYTHONPATH=../sdk/python/sdk` to use the repository SDK.
- `uv run --frozen ruff check tests/test_prisma_dir.py` and `git diff --check`: passed.
- Full Prisma suite: 568 passed, 2 failed. Both failures reproduce on unchanged upstream: `test_open_terminal_agent_mode` depends on an unavailable `claude` executable, and `test_list_branches` lacks a Git identity in its temporary repository. The branch test passes when author/committer environment variables are supplied.

### DIFF
--- a/prisma/src/luml_prisma/data/guide.md
+++ b/prisma/src/luml_prisma/data/guide.md
@@ -190,17 +190,17 @@ After training and saving the model, use `link_to_model()` to embed experiment d
 
 ```python
 from pathlib import Path
-from luml import ExperimentTracker
+from luml.experiments.tracker import ExperimentTracker
 from luml.integrations.sklearn import save_sklearn
 
 tracker = ExperimentTracker(f"sqlite://{Path.home()}/.luml/experiments")
 exp_id = tracker.start_experiment(name="my-experiment")
 # ... training ...
-tracker.log_metrics({"accuracy": 0.92}, step=0)
+tracker.log_dynamic("accuracy", 0.92, step=0)
 
 model_ref = save_sklearn(model, X_train, path=".prisma/artifact.luml")
 tracker.link_to_model(model_ref)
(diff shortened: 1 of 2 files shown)


codie-dfs:
codie started reviewing commit `6e4026b` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `6e4026b` (requested by @OKUA1)

This PR updates the bundled Prisma guide to use the supported `ExperimentTracker` import and lifecycle APIs, then adds a guard against unknown tracker methods. The revised examples match the current API; no correctness or security issues were found.

### Missing test coverage

- `prisma/tests/test_prisma_dir.py:139` validates method names only. It would still pass if the guide regressed to the unsupported import path or used valid methods with invalid arguments. At minimum, assert the documented import path; ideally, validate extracted calls against their signatures.

OKUA1 (approved):

PR #652, opened 2026-09-10
fix/LM-466-remove-satellite-click-affordance

orbie-dfs:
Removes the pointer cursor and hover highlight from non-clickable satellite cards so they no longer imply a details action. Adds a regression test for the card’s non-interactive presentation.


codie-dfs:
codie started reviewing commit `f630db4` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `f630db4` (requested by @OKUA1)

This PR removes the pointer cursor and hover styling from non-clickable satellite cards. The CSS change is appropriate, but the new regression test does not actually validate it.

### Smaller problems

- **Ineffective CSS regression test** — `frontend/src/components/satellites/SatellitesCard.test.ts:32`: The jsdom test configuration does not enable SFC CSS processing, so `getComputedStyle` sees the default cursor even if `cursor: pointer` is restored. The tag-name assertion also passed before this change. Use a browser/Playwright test or explicitly load the component stylesheet so the test fails against the previous implementation.

### Missing test coverage

The removed `:hover` background behavior is not covered. A browser-level test should hover the card and verify both the cursor and background remain non-interactive.

codie-dfs:
codie started reviewing commit `ab71c3c` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `ab71c3c` (requested by @olha-poliuliakkh)

This PR removes the misleading pointer cursor, hover highlight, and transition from non-clickable satellite cards. The new Playwright tests cover both cursor and hover behavior.

No correctness, security, or test-coverage issues found.

limanski-valera (approved):

PR #653, opened 2026-09-10
fix/LM-490-hide-empty-attachments

orbie-dfs:
Hide the Attachments tab only after a bounded, validated index read confirms there are no usable files. Reuse the parsed index in the attachment provider, retain a retry path on inspection errors, and ignore stale navigation results. Adds regression coverage for visibility, failures, malformed ranges, and navigation races.

### DIFF
--- a/extras/js/packages/attachments/src/models/TarAttachmentsProvider.ts
+++ b/extras/js/packages/attachments/src/models/TarAttachmentsProvider.ts
@@ -19,6 +19,7 @@ interface ModelDownloader {
 export interface TarAttachmentsProviderConfig {
   downloader: ModelDownloader
   fileIndex: FileIndex
+  attachmentsIndex?: FileIndex
   findAttachmentsTarPath: (fileIndex: FileIndex) => string | undefined
   findAttachmentsIndexPath: (fileIndex: FileIndex) => string | undefined
 }
@@ -28,6 +29,7 @@ export class TarAttachmentsProvider implements ModelAttachmentsProvider {
   private fileIndex: FileIndex
   private findAttachmentsTarPath: (fileIndex: FileIndex) => string | undefined
   private findAttachmentsIndexPath: (fileIndex: FileIndex) => string | undefined
+  private initialAttachmentsIndex?: FileIndex
 
   private tree: FileNode[] = []
   private tarBaseOffset: number = 0
@@ -38,6 +40,7 @@ export class TarAttachmentsProvider implements ModelAttachmentsProvider {
     this.fileIndex = config.fileIndex
     this.findAttachmentsTarPath = config.findAttachmentsTarPath
     this.findAttachmentsIndexPath = config.findAttachmentsIndexPath
+    this.initialAttachmentsIndex = config.attachmentsIndex
   }
 
   async init(): Promise<void> {
@@ -46,7 +49,9 @@ export class TarAttachmentsProvider implements ModelAttachmentsProvider {
       return
     }
 
-    const indexData = await this.downloader.getFileFromBucket<FileIndex>(this.fileIndex, indexPath)
+    const indexData =
+      this.initialAttachmentsIndex ??
+      (await this.downloader.getFileFromBucket<FileIndex>(this.fileIndex, indexPath))
 
     const tarPath = this.findAttachmentsTarPath(this.fileIndex)
     if (!tarPath) {
--- a/frontend/src/components/orbits/tabs/registry/collection/artifact/ArtifactTabs.vue
+++ b/frontend/src/components/orbits/tabs/registry/collection/artifact/ArtifactTabs.vue
@@ -40,7 +40,6 @@ type Props = {
 
   cardDisabled: boolean
   experimentSnapshotDisabled: boolean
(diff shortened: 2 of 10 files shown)


codie-dfs:
codie started reviewing commit `dcc317a` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `dcc317a` (requested by @olha-poliuliakkh)

This PR hides the Attachments tab unless `FnnxService.hasAttachments` reports usable content, replacing the previous disabled-tab behavior. It adds unit tests for tab visibility and basic archive/index presence.

## Correctness bugs

- **P2 — Serialized index size is not a reliable attachment check** (`frontend/src/lib/fnnx/FnnxService.ts:213`): Any index larger than the exact two-byte `{}` is considered nonempty, but the provider ignores directory entries and zero-byte files (`extras/js/packages/attachments/src/models/TarAttachmentsProvider.ts:103`). Older SDK indexes include a root `attachments/` directory even with no files, and zero-byte attachments are currently accepted, so both cases still display a tab backed by an empty provider. Parse the index using the provider’s file predicate, or persist an explicit usable-file count instead of inferring it from serialized length.

## Missing test coverage

- `frontend/src/lib/fnnx/__tests__/FnnxService.test.ts:64` only tests an exact two-byte empty index. Add regression cases for directory-only indexes and zero-byte entries, ideally including page-level tab visibility.

codie-dfs:
codie started reviewing commit `63fc368` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `63fc368` (requested by @olha-poliuliakkh)

This PR hides the Attachments tab unless the embedded attachment index contains a usable file. It adds unit coverage for the predicate and tab rendering, while making availability depend on an asynchronous storage read.

### Correctness bugs

- **Do not treat an inspection failure as an empty index** — `frontend/src/pages/collection/artifact/index.vue:138`. Availability is set to `false` before requesting the index, and the catch path at line 156 only shows a toast. A transient download-URL or range-request failure therefore hides valid attachments until the page is reloaded. Model loading, confirmed-empty, available, and error states separately; only hide the tab after successfully parsing an empty index, and preserve an in-page retry path on errors.

### Security issues

- **Bound the eager index download** — `frontend/src/pages/collection/artifact/index.vue:151`. The uploaded artifact controls the range recorded for `attachments.index.json`, and this code automatically downloads and JSON-parses that entire range whenever another user opens its details. A malformed or deliberately oversized range can consume substantial memory and freeze the browser. Validate the range against the artifact size and impose a reasonable index-size limit before fetching, or expose an attachment count from a trusted backend inspection.

### Smaller problems

- **Avoid downloading the index twice** — `frontend/src/pages/collection/artifact/index.vue:149`. After this check fetches a download URL and the index, `frontend/src/pages/collection/artifact/AttachmentsView.vue:30` repeats both requests when the tab opens; direct attachment links perform them concurrently. Cache/pass the parsed index or initialize and share the attachment provider.

### Missing test coverage

- The new tests only exercise `hasAttachments()` and a supplied visibility prop. Add page-level tests covering successful non-empty and empty index reads, URL/index failures, and stale asynchronous responses after navigating to another artifact.

codie-dfs:
codie started reviewing commit `8ffabcd` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `8ffabcd` (requested by @olha-poliuliakkh)

This PR determines attachment availability from a validated attachment index, caches the parsed index, and hides or redirects empty attachment routes. It also adds inline loading/error states with retry support and tests for index validation and stale loads.

### Correctness bugs

- **Medium — Retrying into an empty result leaves the user on a hidden route.** `frontend/src/pages/collection/artifact/AttachmentsView.vue:49` reloads attachment state but does not redirect when the result is `empty`; the redirect at `frontend/src/pages/collection/artifact/index.vue:213` only runs during the initial artifact load. If an inspection first fails and a retry successfully returns an empty index, the tab disappears while `/attachments` remains active and the view renders nothing. Redirect after retry or watch the status centrally.

- **Medium — The cached downloader can outlive its signed URL.** `frontend/src/stores/artifacts/index.ts:104` obtains the download URL when the artifact page loads and `frontend/src/stores/artifacts/index.ts:120` retains that downloader for later previews. Because these URLs expire, opening attachments after a long-lived page session—or selecting another file later—can permanently fail without requesting a fresh URL. Cache the validated index, but obtain or refresh the downloader when the attachment view/content is accessed.

### Smaller problems

- **The new 1 MiB index limit rejects otherwise valid artifacts.** `frontend/src/stores/artifacts/index.ts:91` introduces a limit that is not enforced by the attachment producer or artifact API. An artifact with many small attachments can legitimately exceed it and previously remained usable. Either establish the same limit at ingestion/creation time with a clear validation error, or avoid rejecting indexes solely because of their serialized size.

### Missing test coverage

- Add coverage for an inspection failure followed by an empty successful retry, asserting navigation away from `/attachments`.
- Add a package-level test for the supplied `attachmentsIndex` path in `TarAttachmentsProvider`, including content range reads, and coverage for refreshing an expired downloader URL.

PR #654, opened 2026-09-10
feat/LM-587 batch artifact deletion

K-Krut:


codie-dfs:
codie started reviewing commit `095e955` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `095e955` (requested by @K-Krut)

This PR adds three-phase batch artifact deletion across the backend, frontend, and Python SDK, including blocker reporting, force deletion, and deployment-status safeguards. The core flow is consistent, but two frontend state gaps break recovery behavior in the artifact editor.

## Correctness bugs

- **Medium — The editor never observes a deletion failure status.** `frontend/src/components/orbits/tabs/registry/collection/artifact/ArtifactEditor.vue:173` decides whether to offer retry/force solely from `props.data.status`, while `frontend/src/stores/artifacts/index.ts:319` only replaces entries in `artifactsList`. The editor receives an older object or a copy, so after a `storage_error`, closing the result dialog and clicking Delete again still opens the normal confirmation instead of the deletion-failed dialog. Propagate the updated status to the editor/current artifact or maintain an effective local status.

- **Medium — A force retry returning `not_found` leaves the detail page showing a deleted artifact.** The store removes and suppresses `not_found` outcomes at `frontend/src/stores/artifacts/index.ts:297`, but `frontend/src/components/orbits/tabs/registry/collection/artifacts-table/ArtifactsDeletionResultDialog.vue:183` emits only IDs in `result.deleted`. If a force confirmation succeeds server-side but its response is lost, retrying returns `not_found`; the dialog closes, but the artifact-detail page never resets `currentArtifact` or navigates away. Treat these attempted `not_found` IDs as gone for parent notification, while continuing to exclude them from success toasts.

## Missing test coverage

- Add an editor test where deletion produces `storage_error`, the result dialog is closed, and Delete is clicked again.
- Add a detail-page test for a lost force-confirmation response followed by a `not_found` retry, asserting that the stale artifact view is cleared.

codie-dfs:
codie started reviewing commit `06a3dc1` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `06a3dc1` (requested by @K-Krut)

This PR adds collection-scoped, two-phase batch artifact deletion with typed failures, retry/force UI flows, and matching synchronous/asynchronous SDK APIs. It also serializes deployment creation against deletion by locking and rechecking artifact status.

### Correctness bugs

- **High — Track linking can race with deletion and leave a track referencing a missing file.** `backend/luml/repositories/artifacts.py:169` determines eligibility from a one-time track snapshot, then releases the artifact lock before the client deletes the bucket object. Because track-entry creation does not recheck the artifact status under the same lock, a link can be inserted after phase one; phase three then reports the track blocker, but the file is already gone. Serialize track-entry creation on the artifact row and reject `pending_deletion`, so either linking or deletion wins before a URL is issued.

### Smaller problems

- `frontend/src/components/orbits/tabs/registry/collection/artifacts-table/ArtifactsDeletionResultDialog.vue:183` only emits deletion events for `result.deleted`. Since `frontend/src/stores/artifacts/index.ts:297` removes `not_found` outcomes from the returned failures, a force attempt that finds the artifact already deleted closes the dialog without notifying the detail page; `currentArtifact` remains rendered for a nonexistent record. Preserve or infer IDs resolved as `not_found` and reset/navigate the detail page for them too.

### Missing test coverage

- Add a concurrency test for track-entry creation versus `request_batch_deletion`, analogous to the deployment race test.
- Add a detail-page test where forced confirmation returns `not_found` and assert that the stale artifact view is cleared.

OKUA1 (approved):

PR #655, opened 2026-09-10
fix/backend concurrency

K-Krut:


codie-dfs:
codie started reviewing commit `343cc80` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `343cc80` (requested by @K-Krut)

This PR moves deletion and quota checks into row-locked repository transactions, adds uniqueness and foreign-key constraints for several races, and introduces concurrency-focused tests. Several invariants remain bypassable because not every writer participates in the new locking protocol, and the blacklist migration can weaken token revocation.

### Correctness bugs

- **High — Artifact deletion can still race with new references.** `backend/luml/repositories/artifacts.py:86` locks the artifact, counts existing deployments and track entries, then merely updates its status. A deployment or track-entry insert that starts after this lock waits for the foreign-key check, then succeeds after the transaction commits because the artifact row still exists; neither creation path locks the artifact and rejects `pending_deletion`. This can leave a deployment or track pointing at storage the client has deleted. Make both reference-creation operations lock and validate the artifact in the same transaction as their insert, or enforce the invariant database-side.

- **Medium — The per-user organization limit remains bypassable.** `backend/luml/repositories/users.py:216` makes the user lock optional, and the public member-add path calls it without a limit at `backend/luml/handlers/organizations.py:388`. Direct additions can therefore exceed the five-organization cap, and can race with an invite acceptance that did lock and count the user. Every organization-membership writer must participate in the user-lock protocol.

- **Medium — Forced stage deletion can lose to a concurrent assignment.** `backend/luml/repositories/tracks.py:191` clears current assignments before locking the stage, while the actual delete is deferred until commit. An entry can acquire the stage between the `UPDATE` and `DELETE`, causing the new `RESTRICT` foreign key to reject even a forced deletion and roll back all unassignments. Coordinate stage assignment and deletion with a consistent stage-row lock order.

### Security issues

- **High — Blacklist deduplication can shorten a refresh token’s revocation lifetime.** `backend/migrations/versions/039_concurrency_guards.py:30` keeps a duplicate based on UUID ordering rather than the greatest `expire_at`, and `backend/luml/repositories/token_blacklist.py:16` similarly leaves an existing shorter expiry unchanged on conflict. Duplicate expirations can differ because logout substitutes the access-token expiry before blacklisting the refresh token at `backend/luml/handlers/auth.py:248`. Cleanup can consequently remove the surviving blacklist row while the refresh token remains valid. Preserve `MAX(expire_at)` during migration and conflicts, and blacklist the refresh token using its own expiry.

### Smaller problems

- The new foreign-key translations discard which constraint failed. `backend/luml/handlers/tracks.py:266` reports every create-entry FK failure as an invalid stage, including a concurrently deleted artifact, track, or user; `backend/luml/repositories/users.py:231` reports every membership FK failure as a missing organization even though the organization was already locked and the missing row may be the user. Match the specific constraint name and translate only that case.

### Missing test coverage

- Race deletion after it owns the artifact lock against both deployment and track-entry creation.
- Race forced stage deletion against a new stage assignment.
- Cover direct member addition at the per-user cap and concurrently with invite acceptance.
- Exercise blacklist duplicates with different expirations through both migration and conflict handling.
- Cover non-stage foreign-key failures in entry creation.

codie-dfs:
codie started reviewing commit `024898d` (requested by @OKUA1).

codie-dfs (commented):
### codie review of `024898d` (requested by @OKUA1)

This PR moves quota, deletion, refresh-token, and track-stage decisions into locked repository transactions and adds supporting database constraints. The approach closes several races, but stage synchronization and blacklist migration still have gaps.

## Correctness bugs

- **Medium — Concurrent stage-list replacements are not serialized.** `backend/luml/repositories/tracks.py:283` locks only existing stage rows. For a track with no stages, two concurrent updates lock nothing; disjoint desired lists can both commit, leaving their union rather than either requested replacement. Lock the parent `TrackOrm` row before reading the stage set, then query stages using a fresh snapshot.

- **Low — Unrelated constraint failures are misclassified.** `backend/luml/handlers/tracks.py:274` maps every unknown non-FK `IntegrityError` to “Artifact is already an entry,” while `backend/luml/repositories/token_blacklist.py:16` treats every integrity failure as a duplicate token. Explicitly match `uq_track_entries_track_id_artifact_id` and `uq_token_black_list_token`; propagate or generically report every other constraint failure.

## Security issues

- **Medium — Existing revoked refresh tokens retain the incorrect shorter expiry.** `backend/luml/handlers/auth.py:241` fixes new logout records, but `backend/migrations/versions/039_concurrency_guards.py:30` only deduplicates existing rows. Rows written before deployment may still expire at the access-token deadline, after which cleanup removes them while the refresh token remains valid and reusable. Backfill expiry from the stored token payload, conservatively retain legacy rows for the maximum refresh lifetime, or invalidate existing sessions during rollout.

## Smaller problems

- `backend/tests/integration/repository/test_concurrency_guards.py:357` and `backend/tests/integration/repository/test_concurrency_guards.py:582` use fixed sleeps to infer that workers reached a lock. Both tests can pass when the tasks simply have not started yet. Use an event, SQL execution hook, or another deterministic barrier before releasing the held row.

## Missing test coverage

- Add concurrent stage synchronization starting from an empty stage set.
- Add a migration regression test for a legacy refresh blacklist row with the shortened expiry.
- Test unknown non-target integrity constraints.
- Race `request_deletion` directly against deployment and track-entry creation, covering both possible winners rather than manually changing artifact status.

OKUA1 (approved):

PR #656, opened 2026-09-11
fix/LM-191-redirect-invite-links

orbie-dfs:
Invitation emails now link to a dedicated invitation center. Protected navigation preserves a validated invitation destination through password, SSO, and registration flows, while the invitation view reports loading and retryable failures. Tests cover routing and invitation actions.

### DIFF
--- a/backend/luml/handlers/organizations.py
+++ b/backend/luml/handlers/organizations.py
@@ -219,7 +219,7 @@ class OrganizationHandler:
                 invite.email if invite else "",
                 get_invited_by_name(invite),
                 get_organization_email_name(invite),
-                config.APP_EMAIL_URL,
+                f"{config.APP_EMAIL_URL.rstrip('/')}/invitations",
             )
         except Exception as error:
             raise EmailDeliveryError(
--- a/backend/tests/unit/handlers/test_organization_invites.py
+++ b/backend/tests/unit/handlers/test_organization_invites.py
@@ -15,6 +15,7 @@ from luml.schemas.organization import (
     UserInvite,
 )
 from luml.schemas.user import CreateUser, UserOut
+from luml.settings import config
 
 handler = OrganizationHandler()
 
@@ -97,7 +98,12 @@ async def test_send_invite(
 
     assert result == mocked_invite
 
-    mock_send_organization_invite_email.assert_called_once()
+    mock_send_organization_invite_email.assert_called_once_with(
+        mocked_invite.email,
+        "",
+        "",
+        f"{config.APP_EMAIL_URL.rstrip('/')}/invitations",
+    )
     mock_create_organization_invite.assert_awaited_once()
 
 
--- a/frontend/src/components/authorization/AuthorizationWrapper.vue
+++ b/frontend/src/components/authorization/AuthorizationWrapper.vue
@@ -48,29 +48,39 @@ import MicrosoftIcon from '@/assets/img/authorization-services/microsoft.svg'
 import { useAuthStore } from '@/stores/auth'
 import { useUserStore } from '@/stores/user'
 import { onBeforeMount, ref } from 'vue'
-import { useRouter } from 'vue-router'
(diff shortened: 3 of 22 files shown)


codie-dfs:
codie started reviewing commit `580e2b1` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `580e2b1` (requested by @olha-poliuliakkh)

This PR routes organization-invitation emails to a new protected invitation center, extracts the existing invitation table into a shared component, and adds guard and integration coverage. The password sign-in path works, but other supported authentication paths lose the invitation destination.

## Correctness bugs

- **Major — SSO login discards the invitation destination.** `frontend/src/router/middlewares/AuthMiddleware.ts:23` stores the return path only in the sign-in query. The Google and Microsoft actions at `frontend/src/components/authorization/AuthorizationWrapper.vue:66` navigate away without persisting it, and the callback always routes home at `frontend/src/components/authorization/AuthorizationWrapper.vue:95`. Persist a validated return path through the OAuth round trip and honor it after callback.

- **Major — registration discards the destination for new invitees.** `backend/luml/handlers/organizations.py:222` now sends every recipient, including emails without an existing account, to an authenticated route. However, the sign-up link at `frontend/src/pages/SignInPage.vue:74` drops the redirect, and signup/email confirmation never restores it. Preserve the invitation destination through registration and verification.

## Smaller problems

- **Minor — loading and failure are presented as “no invitations.”** `frontend/src/components/user/InvitationsList.vue:47` treats the initially empty store as a completed empty response. On direct navigation, invitations are fetched asynchronously by the watcher at `frontend/src/stores/user.ts:70`, which route navigation does not await; failures also leave the same empty state. Track loading/loaded/error state and render an appropriate loader or retry message.

## Missing test coverage

- `frontend/tests/integration/invitations.spec.ts:5` stops after reaching sign-in. Add round-trip coverage for password login, SSO, and new-account registration.
- `frontend/tests/integration/invitations.spec.ts:19` only verifies rendering. Add accept/reject success and failure cases, including the loading/error states.

codie-dfs:
codie started reviewing commit `13ebcc0` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `13ebcc0` (requested by @olha-poliuliakkh)

This PR routes invitation emails to a new authenticated invitation center, adds loading/retry states, and preserves safe post-auth destinations across the main happy paths. Two common authentication paths still lose that destination, and the integration setup leaves dependent requests unmocked.

## Correctness bugs

- **Medium:** OAuth denial or callback failure loses the invitation destination. Providers return failures without a `code`, so `AuthorizationWrapper` remains on `/sign-in` without a `redirect` query; password sign-in then deletes the stored destination and routes home at `frontend/src/pages/SignInPage.vue:127`. Retrying SSO also removes it at `frontend/src/components/authorization/AuthorizationWrapper.vue:68`. Retain the stored redirect through provider errors and use it as a fallback after successful password authentication.

- **Medium:** Registration continuation exists only in `localStorage` (`frontend/src/pages/SignUpPage.vue:158`) and is read from there after confirmation (`frontend/src/pages/EmailConfirmedPage.vue:24`). If the verification link opens on another device, browser profile, or mail-client webview, “Go to account” cannot preserve `/invitations`. Carry an opaque, validated continuation through the confirmation link or server-side confirmation state.

## Smaller problems

- `mockAuthenticatedInvitationPage` returns an organization but does not mock the consequent organization-orbits request at `frontend/tests/integration/invitations.spec.ts:15`. Organization initialization automatically makes that request, so these tests can contact the configured backend and become timing-dependent. Mock the follow-up request, preferably with an empty orbit list.

## Missing test coverage

- Add OAuth denial/code-exchange-failure tests followed by password sign-in or an SSO retry.
- Add a confirmation-flow test starting with empty browser storage to cover verification in a separate context.

K-Krut (approved):

PR #657, opened 2026-09-11
fix: dynamic attributes cant be set as null when update deployment 

olha-poliuliakkh:
update_deployment_details converted the inbound payload with DeploymentDetailsUpdate.model_validate(data.model_dump(mode="json")). mode="json" was there only to turn the dynamic_attributes_secrets UUID values into strings, but model_dump() also emits every unset field at its default, and the following model_validate() marks them all as explicitly set. That defeated the exclude_unset=True the repository relies on, so any PATCH wrote None over description, schemas, error_message, tags and dynamic_attributes_secrets.

Nothing caught the bad writes. The plain columns are nullable, and for the JSONB ones SQLAlchemy stores None as the JSON value `null` rather than SQL NULL, so their NOT NULL constraints passed. The corruption only surfaced on read: Deployment.dynamic_attributes_secrets is dict[str, str], and JSON null fails validation, turning GET /deployments into a 500.

Pass exclude_unset=True through the conversion, add migration 039 to backfill the JSONB columns holding null and constrain them to objects so a future non-object write fails loudly, and cover the partial update with a test. The existing test could not catch this: it compares pydantic models, and __eq__ ignores model_fields_set.


codie-dfs:
codie started reviewing commit `fb0a09c` (requested by @olha-poliuliakkh).

codie-dfs (commented):
### codie review of `fb0a09c` (requested by @olha-poliuliakkh)

This PR preserves omitted PATCH fields through schema conversion and adds a migration to repair and constrain deployment JSONB columns. The propagation fix is sound, but the migration can discard unrelated data and its constraint conflicts with an accepted request shape.

### Correctness bugs

- **High — `backend/migrations/versions/039_repair_deployment_null_json_fields.py:42`:** The backfill replaces every non-object value—including arrays, strings, numbers, and booleans—with `{}`. The reported corruption only requires repairing SQL/JSON nulls; silently erasing other shapes can destroy recoverable environment, secret-reference, or satellite configuration data. Restrict the update to null values and either explicitly migrate or fail on other unexpected types.

- **Medium — `backend/luml/handlers/deployments.py:322`:** `exclude_unset=True` correctly retains explicitly supplied values, including `dynamic_attributes_secrets: null`, which the request schema accepts at `backend/luml/schemas/deployment.py:127`. The repository then writes `None`, and the new check at `backend/migrations/versions/039_repair_deployment_null_json_fields.py:55` raises an unhandled integrity error, producing a 500. Reject explicit null during request validation or normalize it to `{}`.

### Missing test coverage

- `backend/tests/unit/handlers/test_deployments.py:1124` mocks the repository, so it does not verify that a partial PATCH preserves populated columns after commit and reload.
- Add a migration test upgrading revision 038 data containing JSON null, verifying the backfill and constraint, plus endpoint coverage for explicit null.

codie-dfs:
codie started reviewing commit `1133f58` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `1133f58` (requested by @K-Krut)

This change preserves omitted fields during deployment PATCH requests, defines explicit-null behavior, and repairs legacy JSON `null` values. It also introduces database constraints requiring four deployment JSONB columns to contain objects.

### Correctness / availability

- **Medium:** `backend/migrations/versions/040_repair_deployment_null_json_fields.py:58` adds and validates four CHECK constraints in one transaction. PostgreSQL validates each `ADD CHECK` while holding an `ACCESS EXCLUSIVE` lock, so the first lock remains held across four table scans and blocks deployment reads and writes. Install the constraints as `NOT VALID` in a short transaction, then validate them in a separate migration/transaction.

### Smaller problems

- `backend/luml/schemas/deployment.py:131` rejects explicit `null` for `name` and `monitoring_mode`, but their `str | None` / `MonitoringMode | None` annotations still advertise `null` as valid in OpenAPI. Generated clients can therefore send schema-valid requests that receive 422 responses. Update the input JSON schema to mark these fields non-nullable when present.

- `backend/migrations/versions/040_repair_deployment_null_json_fields.py:58` adds constraints that are absent from `DeploymentOrm.__table_args__`. Since `Base.metadata` is Alembic’s target metadata, schema metadata no longer describes the database invariants. Add matching `CheckConstraint` entries to the ORM model.

### Missing test coverage

- `backend/tests/integration/test_migration_040.py:17` hardcodes only `dynamic_attributes_secrets`; the repair and constraints for `env_variables_secrets`, `env_variables`, and `satellite_parameters` are untested. Parameterize the migration tests across all four columns, and add an OpenAPI assertion for the new non-null request contract.

K-Krut (approved):

PR #658, opened 2026-09-14
fix: opening a collection from the registry no longer shows the 404 page

K-Krut:
Vue mounts the collection page before it unmounts the registry view, whose reset() bumped the request counter and dropped the page's own response. reset() no longer invalidates requests; only the collection page does.


codie-dfs:
codie started reviewing commit `302fdf1` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `302fdf1` (requested by @K-Krut)

This changes `collections.reset()` so a collection request started during the registry-to-collection route swap can still complete. The added store tests cover preserving that response while retaining explicit cancellation through `resetCurrentCollection()`.

No correctness or security issues found.

### Missing test coverage

- `frontend/src/stores/__tests__/collections.test.ts:54` manually reproduces the lifecycle ordering but does not exercise the actual router transition. Consider adding a component/router test that navigates from the registry to a collection and verifies collection content appears instead of `Ui404` after the deferred request resolves.

OKUA1 (approved):

PR #659, opened 2026-09-14
fix: alembic reads POSTGRESQL_DSN from .env again, deploy prints the …

K-Krut:
…applied revision


codie-dfs:
codie started reviewing commit `5843007` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `5843007` (requested by @K-Krut)

This PR switches Alembic’s DSN lookup to a minimal `BaseSettings` model so CLI migrations can load `backend/.env`, while bypassing settings when a connection is injected. The deployment workflow now logs the database revision after upgrading.

### Correctness

- **Medium:** `backend/migrations/env.py:35` unconditionally requires `POSTGRESQL_DSN` whenever Alembic creates its own engine, removing the previous fallback to `config.get_main_option("sqlalchemy.url")`. Custom or programmatic Alembic configurations that provide `sqlalchemy.url` without an environment variable will now fail with a settings validation error. Retain the configured Alembic URL as a fallback when no environment or dotenv value is available.

### Missing test coverage

- `backend/migrations/env.py:24` has no regression tests covering DSN loading from `.env`, environment-over-dotenv precedence, or the configured-URL fallback. Add focused tests for these resolution paths and for the injected-connection path not requiring settings.

OKUA1 (approved):

PR #667, opened 2026-09-17
fix/LM-686-account-deletion-track-version

orbie-dfs:
Allows users to delete their accounts after creating track versions. Track entries are preserved while their creator attribution is cleared when the user is deleted, with migration and integration coverage.


codie-dfs:
codie started reviewing commit `39fb377` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `39fb377` (requested by @K-Krut)

This PR preserves track entries when their creator is deleted by making `added_by` nullable and applying `ON DELETE SET NULL`. The integration test verifies the database and repository behavior.

### Correctness

- **Medium:** `backend/luml/schemas/tracks.py:112` allows entry endpoints to return `added_by: null`, but `sdk/python/api/luml_api/_types.py:428` still requires a UUID. SDK get/list/update methods validate responses with Pydantic, so accessing a retained entry will raise `ValidationError`. `frontend/src/lib/api/orbit-tracks/interfaces.ts:54` also retains the incorrect non-null contract. Update both client types to nullable.

### Missing test coverage

- `backend/tests/integration/repository/test_user.py:111` covers repository retrieval only. Add an SDK/API regression test that parses or lists an entry with `added_by: null`; this would catch the incompatible client model above.

K-Krut (approved):

PR #671, opened 2026-09-18
fix/LM-726-prevent-self-account-lockout

orbie-dfs:
Removes privileged account fields from the self-service profile update schema and rejects unknown properties so attempts to set them return 422. Adds endpoint coverage for protected and valid profile fields.


codie-dfs:
codie started reviewing commit `b3dd537` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `b3dd537` (requested by @K-Krut)

This PR prevents profile updates from modifying protected account fields while preserving those fields for internal update flows. It adds endpoint tests covering rejection of protected fields and acceptance of normal profile fields.

No correctness, security, or test-coverage issues found.

K-Krut (approved):

PR #672, opened 2026-09-18
fix/LM-730-confirm-artifact-deletion-deployment-check

orbie-dfs:
Checks deployment references again when confirming a single artifact deletion, so deployments created after delete URL issuance return the established 409 conflict instead of a database error. Adds regression coverage that verifies the response and prevents the delete attempt.


codie-dfs:
codie started reviewing commit `f9db6bb` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `f9db6bb` (requested by @K-Krut)

This change prevents confirmation of deletion for artifacts with deployments by returning the existing 409 conflict, with regression coverage for blocked and deployment-free cases. The deployment relationship is eagerly loaded and the guard aligns with existing deletion semantics.

No findings.

K-Krut (approved):

PR #673, opened 2026-09-19
fix/LM-732-signup-transaction

orbie-dfs:
Keeps signup persistence inside a transaction until activation email delivery succeeds, so a delivery failure rolls back the user, organization, and membership together.

Adds handler and repository coverage for the failure path.


codie-dfs:
codie started reviewing commit `e9d6932` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `e9d6932` (requested by @K-Krut)

This PR wraps user, organization, and membership creation in one transaction so activation-email failures roll back signup data. It also preserves the existing `create_user()` API and adds rollback-focused tests.

## Correctness bugs

- **Medium — The activation email is sent before the signup transaction commits.** `backend/luml/handlers/auth.py:155` invokes SendGrid while the transaction opened at `backend/luml/repositories/users.py:72` remains uncommitted. If commit fails or the request is cancelled after SendGrid accepts the message, the user receives a confirmation link for an account that does not exist; the transaction and database connection also remain open throughout the external request. Commit durable state before dispatching the email, preferably through a transactional outbox or pending-signup workflow.

## Missing test coverage

- No test covers successful email dispatch followed by failure during context exit/commit. Add coverage for that ordering so an email cannot become externally visible before the signup state is durable.

PR #675, opened 2026-09-19
fix/LM-740-undeploy-before-force-delete

orbie-dfs:
Force deletion now returns the existing 409 artifact-in-use error when deployments still reference the artifact. This preserves deployment records and satellite containers until they are undeployed normally, preventing running containers from becoming invisible to LUML.


codie-dfs:
codie started reviewing commit `0b0581d` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `0b0581d` (requested by @K-Krut)

This PR prevents the single-artifact force-delete path from deleting deployments, returning a conflict instead. The unit tests reflect the new behavior, but the deployment check is not atomic with deletion.

### Correctness

- **Medium — Make deployment rejection atomic** (`backend/luml/handlers/artifacts.py:676`): `get_artifact_details()` completes in a separate session before `_delete_artifact()` starts its transaction. A concurrent deployment can therefore be created after this check passes; the subsequent foreign-key failure becomes `DatabaseConstraintError`, which is not handled by `AppService` and results in HTTP 500 instead of the intended 409. Lock and check the artifact within the deletion transaction, or translate the constraint race to `ArtifactDeployedError`.

### Missing test coverage

The test at `backend/tests/unit/handlers/test_artifacts.py:2802` only supplies a pre-populated mock deployment list. Add a database-backed concurrency test overlapping deployment creation with force deletion, asserting that one operation wins cleanly, references remain valid, and force deletion never returns 500.

codie-dfs:
codie started reviewing commit `8571b00` (requested by @K-Krut).

codie-dfs (commented):
### codie review of `8571b00` (requested by @K-Krut)

This PR adds transactional reference guards, queues satellite undeploy tasks during forced artifact deletion, and makes deletion callbacks idempotent when deployment rows are already gone. The locking closes the reference race, but the forced-undeploy lifecycle and callback scoping still have issues.

## Correctness bugs

- **High — Failed forced undeploys cannot be retried.** `backend/luml/repositories/deployments.py:228` deletes each deployment immediately after merely enqueueing remote work. If container removal fails, `satellite/agent/tasks/undeploy.py:37` marks the task failed, while the controller only polls pending tasks (`satellite/agent/controllers/periodic.py:23`); the deployment and artifact are already gone, so users cannot request deletion again. This can leave an untracked container serving indefinitely. Preserve a retryable deployment/tombstone until undeploy succeeds, or implement durable retry/backoff for these forced tasks.

## Security issues

- **Medium — The callback exposes a cross-satellite deployment-existence oracle.** `backend/luml/handlers/deployments.py:281` performs a global existence query after the satellite-scoped lookup fails. An authenticated satellite therefore receives 404 for another satellite’s deployment ID but 204 for a nonexistent/deleted ID. Return an indistinguishable response for foreign and missing resources, or validate idempotency against a satellite-scoped undeploy task/tombstone.

## Missing test coverage

- `backend/tests/integration/repository/test_artifacts.py:465` stops after verifying that a pending task was created and manually reproduces repository calls rather than exercising the force-delete handler through worker completion. Add coverage for successful task execution, transient container-removal failure followed by retry, and transaction rollback when lineage cleanup fails after tasks and deployments have been staged.

